PR #1, opened 2020-04-08
Add FromMnemonic, change namescope and add github workflow file

paweljakubas:


.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+        plan:
+        - { build: stack, resolver: "--resolver lts-14.25"  }
+        include: []
+
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Install OS Packages
+      uses: mstksg/get-package@v1
+      with:
+        apt-get: libgmp-dev libssl-dev libtinfo-dev libsystemd-dev zlib1g-dev
```
KtorZ, line 22: small note: I need libssl-dev and libsystemd-dev (and probably most of the other except libgmp-dev) for building the consensus and network libraries in cardano-transactions.
I think that here, only `libgmp-dev` and `libssl-dev` are necessary for building cryptonite & cardano-crypto.

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - uses: actions/checkout@v1
+
+    - name: Setup Stack
+      uses: mstksg/setup-stack@v1
+
+    - name: Build & Test
+      run: |
+        stack --no-terminal test --bench --only-dependencies
+        stack --no-terminal test --bench --no-run-benchmarks
```
KtorZ, line 32: I'll suggest also setting up cache steps here.

src/Cardano/Mnemonic.hs
```diff
@@ -263,3 +280,95 @@ mnemonicToText =
...
+    (SomeMnemonic mwa) == (SomeMnemonic mwb) =
+        case typeOf mwa `testEquality` typeOf mwb of
+            Nothing -> False
+            Just Refl -> mwa == mwb
+
+-- | Create a passphrase from a mnemonic sentence. This class enables caller to
+-- parse text list of variable length into mnemonic sentences.
+--
+-- >>> fromMnemonic @'[12,15,18,21] @"generation" ["toilet", "curse", ... ]
+-- Right (Passphrase <ScrubbedBytes>)
```
KtorZ, line 298: That comment is actually outdated. `FromMnemonic` used to return a `Passphrase` but it now returns `SomeMnemonic`.

.github/workflows/main.yml
paweljakubas: indeed, added caching and haddoc

src/Cardano/Mnemonic.hs
paweljakubas: corrected

.github/workflows/main.yml
paweljakubas: yes, `libgmp-dev` and `libssl-dev` are enough

PR #4, opened 2020-04-11
Add byron address derivation

paweljakubas:
- [x] Add first working although coarse Byron impl of address derivation
- [x] add AddressDerivationSpec
- [x] add AddressDerivation.ByronSpec
- [x] clean code

### DIFF
--- a/src/Cardano/AddressDerivation/Byron.hs
+++ b/src/Cardano/AddressDerivation/Byron.hs
@@ -38,15 +38,20 @@ module Cardano.AddressDerivation.Byron
     , mkByronKeyFromMasterKey
 
       -- * Derivation
-    , deriveAccountPrivateKey
-    , deriveAddressPrivateKey
+    , deriveAccountPrivateKeyImpl
+    , deriveAddressPrivateKeyImpl
 
     ) where
 
 import Prelude
 
 import Cardano.AddressDerivation
-    ( Depth (..), DerivationType (..), Index (..) )
+    ( Depth (..)
+    , DerivationType (..)
+    , GenMasterKey (..)
+    , HardDerivation (..)
+    , Index (..)
+    )
 import Cardano.Crypto.Wallet
     ( DerivationScheme (DerivationScheme1)
     , XPrv
@@ -108,6 +113,16 @@ type family DerivationPath (depth :: Depth) :: * where
     -- The address key is generated from the account key and address index.
     DerivationPath 'AddressK =
         (Index 'WholeDomain 'AccountK, Index 'WholeDomain 'AddressK)
+
+instance GenMasterKey ByronKey where
+    type GenMasterKeyFrom ByronKey = SomeMnemonic
+    genMasterKey = generateKeyFromSeed
+
+instance HardDerivation ByronKey where
+    type AddressIndexDerivationType ByronKey = 'WholeDomain
+    deriveAccountPrivateKey = deriveAccountPrivateKeyImpl
+    deriveAddressPrivateKey pwd accXPrv _accStyle ix =
+        deriveAddressPrivateKeyImpl pwd accXPrv ix
 
 {-------------------------------------------------------------------------------
                                  Key generation
@@ -202,12 +217,12 @@ unsafeMkByronKeyFromMasterKey derivationPath masterKey = ByronKey
 
 -- | Derives account private key from the given root private key, using
 -- derivation scheme 1.
-deriveAccountPrivateKey
+deriveAccountPrivateKeyImpl
     :: ScrubbedBytes
     -> ByronKey 'RootK XPrv
     -> Index 'WholeDomain 'AccountK
     -> ByronKey 'AccountK XPrv
-deriveAccountPrivateKey pwd masterKey idx@(Index accIx) = ByronKey
+deriveAccountPrivateKeyImpl pwd masterKey idx@(Index accIx) = ByronKey
     { getKey = deriveXPrv DerivationScheme1 pwd (getKey masterKey) accIx
     , derivationPath = idx
     , payloadPassphrase = payloadPassphrase masterKey
@@ -215,12 +230,12 @@ deriveAccountPrivateKey pwd masterKey idx@(Index accIx) = ByronKey
 
 -- | Derives address private key from the given account private key, using
 -- derivation scheme 1.
-deriveAddressPrivateKey
+deriveAddressPrivateKeyImpl
     :: ScrubbedBytes
     -> ByronKey 'AccountK XPrv
     -> Index 'WholeDomain 'AddressK
     -> ByronKey 'AddressK XPrv
-deriveAddressPrivateKey pwd accountKey idx@(Index addrIx) = ByronKey
+deriveAddressPrivateKeyImpl pwd accountKey idx@(Index addrIx) = ByronKey
     { getKey = deriveXPrv DerivationScheme1 pwd (getKey accountKey) addrIx
     , derivationPath = (derivationPath accountKey, idx)
     , payloadPassphrase = payloadPassphrase accountKey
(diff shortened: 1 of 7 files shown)


src/Cardano/AddressDerivation/Byron.hs
```diff
@@ -0,0 +1,252 @@
...
+minSeedLengthBytes :: Int
+minSeedLengthBytes = 16
+
+-- | Generate a root key from a corresponding seed.
+-- The seed should be at least 16 bytes.
+generateKeyFromSeed
+    :: SomeMnemonic
+    -> ScrubbedBytes
+    -> ByronKey 'RootK XPrv
+generateKeyFromSeed = unsafeGenerateKeyFromSeed ()
```
KtorZ, line 124: This should really implement: https://input-output-hk.github.io/cardano-addresses/haddock/Cardano-AddressDerivation.html#t:GenMasterKey ?

src/Cardano/AddressDerivation/Byron.hs
```diff
@@ -0,0 +1,252 @@
...
+deriveAddressPrivateKey
+    :: ScrubbedBytes
+    -> ByronKey 'AccountK XPrv
+    -> Index 'WholeDomain 'AddressK
+    -> ByronKey 'AddressK XPrv
+deriveAddressPrivateKey pwd accountKey idx@(Index addrIx) = ByronKey
+    { getKey = deriveXPrv DerivationScheme1 pwd (getKey accountKey) addrIx
+    , derivationPath = (derivationPath accountKey, idx)
+    , payloadPassphrase = payloadPassphrase accountKey
+    }
```
KtorZ, line 226: This should implement: https://input-output-hk.github.io/cardano-addresses/haddock/Cardano-AddressDerivation.html#t:HardDerivation, no?

src/Cardano/AddressDerivation/Byron.hs
```diff
@@ -0,0 +1,252 @@
...
+invariant
+    :: String
+        -- ^ The message
+    -> a
+        -- ^ The value to test
+    -> (a -> Bool)
+        -- ^ The predicate
+    -> a
+invariant msg a predicate =
+    if predicate a then a else error msg
```
KtorZ, line 252: What about using: https://hackage.haskell.org/package/base-4.6.0.1/docs/Control-Exception-Base.html#v:assert instead? It's built-in, and raises proper exceptions.

test/Cardano/AddressDerivationSpec.hs
```diff
@@ -0,0 +1,150 @@
...
+instance Arbitrary (Index 'WholeDomain 'AddressK) where
+    shrink _ = []
+    arbitrary = arbitraryBoundedEnum
+
+instance Arbitrary (Index 'WholeDomain 'AccountK) where
+    shrink _ = []
+    arbitrary = arbitraryBoundedEnum
+
+instance Arbitrary SomeMnemonic where
+    arbitrary = SomeMnemonic <$> genMnemonic @12
```
KtorZ, line 124: What do you think of moving all the `Arbitrary` instances inside a `Test.Arbitrary` module that every spec file imports? This project will be small enough and tests done in the same context, so it makes sense to share the instances across the various specs here.

src/Cardano/AddressDerivation/Byron.hs
```diff
@@ -0,0 +1,252 @@
...
+-- Implementation of address derivation for the random scheme, as
+-- implemented by the legacy Cardano wallets.
+--
+-- For full documentation of the key derivation schemes,
+-- see the "Cardano.Crypto.Wallet" module, and the implementation in
+-- <https://github.com/input-output-hk/cardano-crypto/blob/4590efa638397e952a51a8994b5543e4ea3c1ecd/cbits/encrypted_sign.c cardano-crypto>.
+
+module Cardano.AddressDerivation.Byron
+    ( -- * Types
+      ByronKey(..)
```
KtorZ, line 31: What about calling this just `Byron` :thinking: or, `Random`?

src/Cardano/AddressDerivation/Byron.hs
paweljakubas: https://github.com/input-output-hk/cardano-addresses/pull/4/commits/22ce22c2de17a43e34d027ec0eb18933e086d814

src/Cardano/AddressDerivation/Byron.hs
paweljakubas: added in https://github.com/input-output-hk/cardano-addresses/pull/4/commits/22ce22c2de17a43e34d027ec0eb18933e086d814

src/Cardano/AddressDerivation/Byron.hs
paweljakubas: corrected in https://github.com/input-output-hk/cardano-addresses/pull/4/commits/06ab4de0503ee77462630ceb6a7b4f195fb80a75

test/Cardano/AddressDerivationSpec.hs
paweljakubas: done in commits : https://github.com/input-output-hk/cardano-addresses/pull/4/commits/fcb6ee2df022d7fbf40f3c4f016311137f50f0b2 https://github.com/input-output-hk/cardano-addresses/pull/4/commits/6a67c795f811ea3a466b69a8bbbbd1024305824d https://github.com/input-output-hk/cardano-addresses/pull/4/commits/8b5fd880422179f83fe3639831bf5a188b85e56a

src/Cardano/AddressDerivation/Byron.hs
paweljakubas: renamed to Byron in https://github.com/input-output-hk/cardano-addresses/pull/4/commits/4034f1e2c12862632d481703f9577c1b155a479f

src/Cardano/AddressDerivation/Byron.hs
```diff
@@ -0,0 +1,242 @@
...
+      -- * Generation
+    , unsafeGenerateKeyFromSeed
+    , generateKeyFromSeed
+    , minSeedLengthBytes
+    , unsafeMkByronKeyFromMasterKey
+    , mkByronKeyFromMasterKey
+
+      -- * Derivation
+    , deriveAccountPrivateKeyImpl
+    , deriveAddressPrivateKeyImpl
```
KtorZ, line 42: We may review this a bit later but I don't think we should export these two functions. That's why we have the `HardDerivation` interface.

src/Cardano/AddressDerivation/Byron.hs
```diff
@@ -0,0 +1,242 @@
...
+-- hash function (Blake2b) rather than key stretching with PBKDF2.
+--
+-- There are two methods of hashing the seed entropy, for different use cases.
+--
+-- 1. Normal random derivation wallet seeds. The seed entropy is hashed using
+--    Blake2b_256, inside a double CBOR serialization sandwich.
+--
+-- 2. Seeds for redeeming paper wallets. The seed entropy is hashed using
+--    Blake2b_256, without any serialization.
+hashSeed :: ScrubbedBytes -> ScrubbedBytes
```
KtorZ, line 179: Now that I see this one, it feels like a lie to see `ScrubbedBytes` used in type signatures whereas we do a double conversion to strict bytestrings inside the function :upside_down_face: ... It kinda looses a bit the benefits of using `ScrubbedBytes` in the first place. Though we don't have much choice because `encodeBytes` needs a `ByteString`.

We could perhaps adjust the signature to be slightly more honest and be `ByteString -> ByteString` instead.

src/Cardano/AddressDerivation/Byron.hs
paweljakubas: indeed, the same can be applied to `generateFromSeed` as now it is part of `GenMasterKey` interface

src/Cardano/AddressDerivation/Byron.hs
paweljakubas: will try to address this remark in next PR - now do not see how moving to `ByteString -> ByteString` can simplify things. We use `ScrubbedBytes` as passwords, `blake2b256` deals with its, but `cbor` handles `ByteString` :thinking:

PR #8, opened 2020-04-16
Add address payment

paweljakubas:
- [x] Add AddressPayment interface with other bits
- [x] Add Cbor decoders and encoders
- [x] Add CborSpec
- [x] Add Icarus impl of AddressPayment
- [x] Add Byron impl of AddressPayment
- [x] Add more tests to IcarusSpec

### DIFF
--- a/cardano-addresses.cabal
+++ b/cardano-addresses.cabal
@@ -94,6 +94,9 @@ test-suite unit
   build-depends:
       QuickCheck
     , base >=4.7 && <5
+    , base58-bytestring
+    , bech32
+    , bech32-th
     , bytestring
     , cardano-addresses
     , cardano-crypto
--- a/cardano-addresses.cabal
+++ b/cardano-addresses.cabal
@@ -42,8 +42,8 @@ library
   exposed-modules:
       Cardano.AddressDerivation
       Cardano.AddressDerivation.Byron
-      Cardano.AddressDerivation.Cbor
       Cardano.AddressDerivation.Icarus
+      Cardano.Codec.Cbor
       Cardano.Mnemonic
   other-modules:
       Paths_cardano_addresses
@@ -51,7 +51,10 @@ library
       src
   build-depends:
       base >=4.7 && <5
+    , base58-bytestring
     , basement
+    , bech32
+    , bech32-th
     , bytestring
     , cardano-crypto
     , cborg
@@ -83,8 +86,8 @@ test-suite unit
   other-modules:
       Cardano.AddressDerivationSpec
       Cardano.AddressDerivation.ByronSpec
-      Cardano.AddressDerivation.CborSpec
       Cardano.AddressDerivation.IcarusSpec
+      Cardano.Codec.CborSpec
       Cardano.MnemonicSpec
       Paths_cardano_addresses
       Test.Arbitrary
@@ -94,9 +97,6 @@ test-suite unit
   build-depends:
       QuickCheck
     , base >=4.7 && <5
-    , base58-bytestring
-    , bech32
-    , bech32-th
     , bytestring
     , cardano-addresses
     , cardano-crypto
--- a/cardano-addresses.cabal
+++ b/cardano-addresses.cabal
@@ -56,6 +56,7 @@ library
     , cborg
     , cryptonite
     , deepseq
+    , digest
     , exceptions
     , fmt
     , memory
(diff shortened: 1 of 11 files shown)


src/Cardano/AddressDerivation.hs
```diff
@@ -202,3 +223,48 @@ class GenMasterKey (key :: Depth -> * -> *) where
...
+    { unAddress :: ByteString
+    } deriving (Show, Generic, Eq, Ord)
+
+instance NFData Address
+
+-- | Available network options.
+data NetworkDiscriminant = Mainnet | Testnet Nat
+
+class NetworkDiscriminantVal (n :: NetworkDiscriminant) where
+    networkDiscriminantVal :: Text
```
KtorZ, line 239: I think this can be removed :+1:

src/Cardano/AddressDerivation.hs
```diff
@@ -202,3 +223,48 @@ class GenMasterKey (key :: Depth -> * -> *) where
...
+instance NetworkDiscriminantVal 'Mainnet where
+    networkDiscriminantVal =
+        "mainnet"
+
+instance KnownNat pm => NetworkDiscriminantVal ('Testnet pm) where
+    networkDiscriminantVal =
+        "testnet (" <> T.pack (show $ natVal $ Proxy @pm) <> ")"
+
+-- | Magic constant associated to a given network
+newtype ProtocolMagic = ProtocolMagic { getProtocolMagic :: Int32 }
```
KtorZ, line 250: Note sure why we are using an `Int32` here rather than a `Word32` :thinking:  ?

cardano-addresses.cabal
```diff
@@ -42,6 +42,7 @@ library
   exposed-modules:
       Cardano.AddressDerivation
       Cardano.AddressDerivation.Byron
+      Cardano.AddressDerivation.Cbor
```
KtorZ, line 45: Feels weird to have this module here because, it's not specific of "AddressDerivation". What about something like:

```
Cardano.Codec.Cbor
```

?

src/Cardano/AddressDerivation.hs
```diff
@@ -202,3 +223,48 @@ class GenMasterKey (key :: Depth -> * -> *) where
...
+
+-- | Address is mere wrapper around a ByteString and represents an encoded address.
+newtype Address = Address
+    { unAddress :: ByteString
+    } deriving (Show, Generic, Eq, Ord)
+
+instance NFData Address
+
+-- | Available network options.
+data NetworkDiscriminant = Mainnet | Testnet Nat
```
KtorZ, line 236: From this module perspective though, the fact that something is mainnet or testnet is anecdotal. What really matters is whether addresses requires magic or don't. 

I'd suggest to reflect this as `RequiresMagic | RequiresNoMagic`.

src/Cardano/AddressDerivation.hs
paweljakubas: indeed, will remove

src/Cardano/AddressDerivation.hs
paweljakubas: yes, changed to Word32 and cbor encoders/decoders also respectively

cardano-addresses.cabal
paweljakubas: will change as you request

src/Cardano/AddressDerivation.hs
```diff
@@ -202,3 +223,48 @@ class GenMasterKey (key :: Depth -> * -> *) where
...
+-- | Hard-coded protocol magic for the Byron MainNet
+mainnetMagic :: ProtocolMagic
+mainnetMagic =  ProtocolMagic 764824073
+
+-- | Derive testnet magic from a type-level Nat
+testnetMagic :: forall pm. KnownNat pm => ProtocolMagic
+testnetMagic = ProtocolMagic $ fromIntegral $ natVal $ Proxy @pm
+
+-- | Encoding of addresses for certain key types and backend targets.
+class PaymentAddress (network :: NetworkDiscriminant) key where
```
KtorZ, line 262: I think we could better organize this instead of putting everything in `Cardano.AddressDerivation`. What about having:


```
src/
└── Cardano
    ├── Address
    │   ├── Derivation.hs
    │   ├── Derivation
    │   │    ├── Byron.hs
    │   │    └── Icarus.hs
    │   └── Payment.hs
    └── Mnemonic.hs
```

KtorZ: And putting bits relevant to the construction of Payment address in `Cardano.Address.Payment`

PR #9, opened 2020-04-17
Package structure re-organization

KtorZ:
- 2b65818d41751949ed2bbd5561d92a8eb8b84487
  :round_pushpin: **simply stack snapshot, use stack LTS directly**
  There's no need to pull in all the Cardano eco-system here because, we don't actually
use any of it (apart from cardano-crypto). So, the snapshot can actually be much simpler

- c3eadbef8c68cb8b8db393611f004a48dfae1f34
  :round_pushpin: **switch to 'package.yaml' instead of raw cabal**
  hpack is integrated by default with stack and is very handy for managing haskell
projects, less verbose and awkward than the cabal format.

- b855f3f9ecccb1fa6ee92e0317ca32e4ce7710bb
  :round_pushpin: **rename and re-organize modules**
    src/
  └── Cardano
      ├── Address
      │   ├── Derivation.hs
      │   └── Style
      │       ├── Byron.hs
      │       └── Icarus.hs
      ├── Address.hs
      ├── Codec
      │   └── Cbor.hs
      └── Mnemonic.hs


KtorZ:
> We are going to resign from cabal and work on packages.yaml in all repos, right?

This plays well with both stack and nix, so, as long as stack remains our main build tool, yes.

PR #10, opened 2020-04-17
Module adjustments / simplifications

KtorZ:
- 4ae694ee10e25a4860537bd432cb5ded78b87c39
  :round_pushpin: **move 'NetworkDiscriminant' down to the term level**
  There's no need to have this as a type-level constraint here. It is just adding unecessary complexity

- 8aa7abc3299f91c2bb87e6d1260c059f7b1cebcc
  :round_pushpin: **re-locate and rename/split 'addressToText' to 'base58' and 'bech32'**
  
- e77ecaa351b09f935ada68ed6f6b03bf9ad87234
  :round_pushpin: **simplify extensions' set for 'Cardano.Address.Derivation'**

### DIFF
--- a/src/Cardano/Address.hs
+++ b/src/Cardano/Address.hs
@@ -1,10 +1,6 @@
-{-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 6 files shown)


src/Cardano/Address.hs
```diff
@@ -28,42 +36,81 @@ import Control.DeepSeq
...
     -- Note that 'paymentAddress' is ambiguous and requires therefore a type
     -- application.
-    paymentAddress :: key 'AddressK XPub -> Address
-
-instance NFData Address
+    paymentAddress :: NetworkDiscriminant -> key 'AddressK XPub -> Address
 
 -- | Available network options.
-data NetworkDiscriminant = Mainnet | Testnet Nat
+data NetworkDiscriminant
```
paweljakubas, line 78: very nice!

PR #11, opened 2020-04-17
First documentation adjustments:

KtorZ:
# Overview

- 21fb4567c2a591c34624c97471bddc43589f2658
  :round_pushpin: **revise module export hierarchy in Cardano.Mnemonic**
  Put the most useful stuff on the top, and the more internal ones on the bottom.

- e6a373c9f93211f2122a9e1f21de88469c50642a
  :round_pushpin: **hide the internal constructor for 'Address'**
  This only adds noise and may encourage people to construct their own address from
scratch. So, we only publicly expose an opaque 'Address' type and keep the internals
hidden. Still, a careful and curious reader looking at the source may decide to do
otherwise, which is fine.

- 9189c7e6a2477ad4cd0e32a90307b3ee6638cdc8
  :round_pushpin: **have function for doing address encoding / decoding inside 'Cardano.Address'**
  I had to remove some circular dependencies with Cardano.Codec.Cbor in order to do that and therefore, resorted to use more primitives in this low-level module

- afce3ad8384b7b645b52df5fcce833bae946b7fc
  :round_pushpin: **cleanup unnecessary extensions from Cardano.MnemonicSpec**
  

# Comments

TODO: add roundtrip tests for encoding/decoding bech32 and base58 addresses (I am on it).

### DIFF
--- a/src/Cardano/Address.hs
+++ b/src/Cardano/Address.hs
@@ -152,7 +152,7 @@ stagingDiscriminant = RequiresNoMagic
 testnetDiscriminant :: NetworkDiscriminant
 testnetDiscriminant = RequiresMagic testnetMagic
 
--- | Magic constant associated to a given network. This is mainly use in two
+-- | Magic constant associated with a given network. This is mainly used in two
 -- places:
 --
 -- (1) In 'Address' payloads, to discriminate addresses between networks.
--- a/src/Cardano/Address.hs
+++ b/src/Cardano/Address.hs
@@ -14,9 +14,11 @@ module Cardano.Address
     , PaymentAddress (..)
     , unsafeMkAddress
 
-      -- * Conversion To Text
+      -- * Conversion From / To Text
     , base58
+    , fromBase58
     , bech32
+    , fromBech32
 
       -- * Network Discrimination
     , NetworkDiscriminant (..)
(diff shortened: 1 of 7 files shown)


src/Cardano/Address.hs
```diff
@@ -1,17 +1,24 @@
...
 {-# LANGUAGE QuasiQuotes #-}
 
+{-# OPTIONS_HADDOCK prune #-}
+
 module Cardano.Address
     ( -- * Address
-      Address (..)
+      Address
     , PaymentAddress (..)
+    , unsafeMkAddress
```
paweljakubas, line 15: wonder if we want to expose it when we have from/to text

src/Cardano/Address.hs
```diff
@@ -49,68 +62,122 @@ import qualified Codec.Binary.Bech32 as Bech32
...
 stagingDiscriminant :: NetworkDiscriminant
 stagingDiscriminant = RequiresNoMagic
 
 -- | Required network discrimination settings for testnet
+--
+-- @since 1.0.0
 testnetDiscriminant :: NetworkDiscriminant
 testnetDiscriminant = RequiresMagic testnetMagic
 
 -- | Magic constant associated to a given network. This is mainly use in two
```
paweljakubas, line 155:
```suggestion
-- | Magic constant associated with a given network. This is mainly used in two
```

src/Cardano/Address.hs
paweljakubas: ah, ok. it is used in style extensively

src/Cardano/Codec/Cbor.hs
```diff
@@ -41,10 +43,6 @@ module Cardano.Codec.Cbor
 
 import Prelude
 
-import Cardano.Address
-    ( Address (..), ProtocolMagic (..) )
-import Cardano.Address.Derivation
-    ( Depth (..), DerivationType (..), Index (..) )
```
paweljakubas, line 47: very good decision to make Cbor agnostic of our types and use Word32, ByteString

src/Cardano/Address.hs
paweljakubas: corrected

src/Cardano/Address.hs
KtorZ: Yes, and it is exposed internally, but not visible from Haddock :)

PR #13, opened 2020-04-17
cardano-crypto wrapper

KtorZ:
- 25857dbfa1b93b28203de29b065d2282d9725838
  :round_pushpin: **provide thin wrapper on top of crypto primitives**
  The cardano-crypto is quite 'old' and has a lot of legacy inherited from past decision. One of them
was the decision of keeping every key 'XPrv' encrypted in-memory which has been said useless by our
cryptographers quite a few times.
It is necessary to encrypt the key when storing it into a more persistent location, but this isn't a
concern for this library. We can therefore simplify quite a few interfaces by removing the question
of encryption passphrases altogether.

Another annoyance of 'cardano-crypto' is that it internally stores and represent XPrv as a combination of:
  prv <> pub <> cc, and all conversion functions are made to and from this internal representation.
This is not very handy for in practice, we are often interested in either prv <> cc or pub <> cc. Thus,
our wrapper provide the necessary bits for correctly handling this conversion.

- 9eb52c494a712822d7495a3e3bf23f58ed3a26ce
  :round_pushpin: **add custom rule to hlint to warn developers if using functions from 'Cardano.Crypto.Wallet' directly**
  
- ed5229e24d9dca7d6b32f7b377708d804a012e9d
  :round_pushpin: **further small documentation improvements in Cardano.Address.Derivation**
  
- a19d6a5eb145ce3d0535f9ef5dfa9f1953d56ac7
  :round_pushpin: **Make 'Byron' and 'Icarus' functors and hide constructors**
  This makes for a slightly better public API, and allows for better maintainability. Users of the modules are
forced to use the provided API instead of constructing the types themselves.

### DIFF
--- a/src/Cardano/Address/Derivation.hs
+++ b/src/Cardano/Address/Derivation.hs
@@ -25,7 +25,7 @@ module Cardano.Address.Derivation
     , HardDerivation (..)
     , SoftDerivation (..)
 
-    -- * Low-Level Cryptography
+    -- * Low-Level Cryptography Primitives
     -- ** XPub
     , XPub
     , xpubFromBytes
@@ -86,7 +86,7 @@ import qualified Crypto.ECC.Edwards25519 as Ed25519
 -- and then basing on it enable address derivation
 
 --
--- Low-Level Cryptography
+-- Low-Level Cryptography Primitives
 --
 
 -- | An opaque type representing an extended private key.
@@ -204,6 +204,8 @@ deriveXPub ds pub (Index ix) =
 -- Generate an XPrv using the legacy method (Byron).
 --
 -- The seed needs to be at least 32 bytes, otherwise an asynchronous error is thrown.
+--
+-- __internal__
 generate
     :: ByteArrayAccess seed
     => seed
@@ -214,6 +216,8 @@ generate seed =
 -- Generate an XPrv using the new method (Icarus).
 --
 -- The seed needs to be at least 16 bytes.
+--
+-- __internal__
 generateNew
     :: (ByteArrayAccess seed, ByteArrayAccess sndFactor)
     => seed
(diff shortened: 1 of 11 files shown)


src/Cardano/Address/Derivation.hs
```diff
@@ -41,12 +73,158 @@ import Fmt
...
+
+-- | Convert an 'XPrv' to a raw 'ByteString'.
+--
+-- @since 1.0.0
+xprvToBytes :: XPrv -> ByteString
+xprvToBytes =
+    stripPub . CC.unXPrv
+  where
+    -- From  < xprv | pub | cc >
+    -- ↳ To  < xprv |     | cc >
```
paweljakubas, line 145: great piece of documentation :100:

src/Cardano/Address/Derivation.hs
```diff
@@ -41,12 +73,158 @@ import Fmt
...
+--
+-- @since 1.0.0
+verify
+    :: ByteArrayAccess msg
+    => XPub
+    -> msg
+    -> XSignature
+    -> Bool
+verify =
+    CC.verify -- re-exported like to allow documenting it.
```
paweljakubas, line 180:
```suggestion
    CC.verify -- re-exported for the sake of documention.
```

src/Cardano/Address/Derivation.hs
```diff
@@ -190,4 +366,14 @@ class GenMasterKey (key :: Depth -> * -> *) where
...
+    genMasterKey :: GenMasterKeyFrom key -> key 'RootK XPrv
+
+
+--
+-- Internals
+--
+
+-- Simple helper to convert 'Either e' to 'Maybe' until we have a slightly
+-- better approach for error.
+eitherToMaybe :: Either e a -> Maybe a
```
paweljakubas, line 403: why not to use `import Data.Either.Extra (eitherToMaybe)` we import already `extra` package

src/Cardano/Address/Style/Byron.hs
```diff
@@ -100,6 +101,7 @@ data Byron (depth :: Depth) key = Byron
 instance (NFData key, NFData (DerivationPath depth)) => NFData (Byron depth key)
 deriving instance (Show key, Show (DerivationPath depth)) => Show (Byron depth key)
 deriving instance (Eq key, Eq (DerivationPath depth)) => Eq (Byron depth key)
+deriving instance (Functor (Byron depth))
```
paweljakubas, line 104: :+1:

src/Cardano/Address/Derivation.hs
paweljakubas: corrected

src/Cardano/Address/Derivation.hs
paweljakubas: done

KtorZ: Ah! I didn't notice, but good point!

Issue #16, opened 2020-04-23
Create Shelley impl of Hard/SoftDerivation and GenMasterKey


paweljakubas:
# Context

We adding support for Shelley addresses in order to work with Shelley-era nodes as implemented in cardano-node

# Decision

We are after `GenMasterKey` and `HardDerivation` / `SoftDerivation` implementations

# Acceptance Criteria

1. `GenMasterKey` interface implemented.
2. `(Hard/Soft)Derivation` interface implemented.
3. Illustrative unit tests are added

---

# Development 

# QA 

The implementation is added in `src/Cardano/Address/Style/Shelley.hs` along with extensive comments. The respective tests are added in `test/Cardano/Address/ShelleySpec.hs`

piotr-iohk:
lgtm.

Question: is ADP-158 only to be implemented in cardano-addresses? (asking as there's no related prs on `cardano-wallet` and `cardano-wallet` does not use `cardano-addresses` atm)

cc: @paweljakubas, @KtorZ

paweljakubas:
After we have it here in `cardano-addresses` we are going to use this library inside `cardano-wallet` and use mnemonics and address derivation

PR #18, opened 2020-04-23
Add shelley key impl part1

paweljakubas:
Issue #16 

1. Introduced `Shelley` key
2. Implemented `GenMasterKey` interface
3. Implemented `Hard/SoftDerivation` interface
4. Added unit tests

### DIFF
--- a/src/Cardano/Address/Style/Shelley.hs
+++ b/src/Cardano/Address/Style/Shelley.hs
@@ -1,0 +1,217 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+{-# OPTIONS_HADDOCK prune #-}
+
+-- |
+-- Copyright: © 2018-2020 IOHK
+-- License: Apache-2.0
+
+module Cardano.Address.Style.Shelley
+    ( -- $overview
+
+      -- * Shelley
+      Shelley
+
+      -- * Accessors
+    , getKey
+
+      -- * Unsafe
+    , liftXPrv
+
+      -- Internals
+    , minSeedLengthBytes
+    , publicKeySize
+    , addrSingleSize
+    , addrGroupedSize
+    ) where
+
+import Prelude
+
+import Cardano.Address.Derivation
+    ( AccountingStyle
+    , Depth (..)
+    , DerivationScheme (..)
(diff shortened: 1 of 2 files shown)


src/Cardano/Address/Style/Shelley.hs
```diff
@@ -0,0 +1,217 @@
...
+
+instance HardDerivation Shelley where
+    type AccountIndexDerivationType Shelley = 'Hardened
+    type AddressIndexDerivationType Shelley = 'Soft
+    type WithAccountStyle Shelley = AccountingStyle
+
+    deriveAccountPrivateKey (Shelley rootXPrv) accIx =
+        let
+            purposeIx =
+                toEnum @(Index 'Hardened _) $ fromEnum purposeIndex
```
Anviking, line 133: Maybe `purposeIndex` and `coinTypeIndex` should already be of type `Index 'Hardened` instead of converting here?

paweljakubas: yeah, golden tests are going to appear in forthcoming PR where `PaymentAddress` interface (and exact structure of address) is implemented. Now, "internal" representation generated from mnemonics, and paths to derivation

paweljakubas:
@Anviking yeah, golden tests are going to appear in forthcoming PR where PaymentAddress interface (and exact structure of address) is implemented. Now, "internal" representation generated from mnemonics, and paths to derivation

src/Cardano/Address/Style/Shelley.hs
paweljakubas: hmm, need to think about it

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -0,0 +1,217 @@
...
+publicKeySize :: Int
+publicKeySize = 32
+
+-- Serialized length in bytes of a Single Address
+addrSingleSize :: Int
+addrSingleSize = 1 + publicKeySize
+
+-- Serialized length in bytes of a Grouped Address
+addrGroupedSize :: Int
+addrGroupedSize = addrSingleSize + publicKeySize
```
KtorZ, line 186: Hmmm. Grouped and Single addresses were Jörmungandr constructs. 

Note that, what is currently called "Shelley" in cardano-wallet should really be called "Jormungandr" ! 
Shelley addresses are actually going to be different. What should land in this module should be the Shelley stuff, not the Jörmungandr ones.

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -0,0 +1,217 @@
...
+-- This can be useful however when serializing / deserializing such a type, or to
+-- speed up test code (and avoid having to do needless derivations from a master
+-- key down to an address key for instance).
+--
+-- @since 1.0.0
+liftXPrv :: XPrv -> Shelley depth XPrv
+liftXPrv = Shelley
+
+instance GenMasterKey Shelley where
+    type SecondFactor Shelley = Maybe SomeMnemonic
```
KtorZ, line 113: SecondFactor should just be a plain `ScrubbedBytes` here. cardano-wallet's API may chose to encode this as a mnemonic, but that's only from an API perspective and shouldn't leak inside this module.

Issue #19, opened 2020-04-23
Add PaymentAddress implementation for Shelley


paweljakubas:
# Context

We adding support for Shelley addresses in order to work with Shelley-era nodes as implemented in cardano-node

# Decision

We are after `PaymentAddress` implementation for Shelley. In Shelley, addresses will have a new format. There exist 9 types of addresses described in the following document https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley/chain-and-ledger/executable-spec/cddl-files/shelley.cddl .We are currently looking to support all of them (as in, being able to recognize all of them), but the wallet and libraries will only make active use of the base address type 0, containing a public spending key hash and a public delegation key hash.
Later, the wallet and libraries will implement support for type 4, pointer addresses.The below is also the reference:
https://github.com/input-output-hk/cardano-ledger-specs/compare/cddl_address_notes#diff-d7073d7dcbeeb9e050157bf7f76d541bR73

# Acceptance Criteria

1. `PaymentAddress` interface should be implemented
2. Golden tests and properties should be added to unit tests suite

---

# Development 

# QA 

a number of things were added : 
1. Shelley impl of current interfaces
2. DelegationAddress and NetworkDescriminant interface and impls (also for Shelley)
3. Tests were added in ShelleySpec and also AddressSpec nad Test.Arbitratry were extended
4. Also numerous comments were added.

piotr-iohk:
LGTM.

PR #20, opened 2020-04-27
Add address impl for shelley

paweljakubas:
Task : #19 

1. Add payment address impl for Shelley - ie., for enterprise addresses
2. Add example of usage
3. Change SndFactor for Shelley and use ScrubbedBytes.
4. Added XPub to ByteString impl not using chain code and using it
5. Added DelegationAddress interface
6. Added DelegationAddress impl for Shelley
7. Introduce HasNetworkDiscriminant for better dealing with network discriminant in both byron/shelley
8. Added roundtrip tests for Shelley

### DIFF
--- a/src/Cardano/Address/Derivation.hs
+++ b/src/Cardano/Address/Derivation.hs
@@ -31,7 +31,8 @@ module Cardano.Address.Derivation
     , XPub
     , xpubFromBytes
     , xpubToBytes
-    , xpubToBytesWithoutChainCode
+    , getPublicKey
+    , getChainCode
 
     -- ** XPrv
     , XPrv
@@ -117,19 +118,26 @@ xpubFromBytes :: ByteString -> Maybe XPub
 xpubFromBytes =
     eitherToMaybe . CC.xpub
 
--- | Convert an 'XPub' to a raw 'ByteString' (public key plus chain code)
+-- | Convert an 'XPub' to a raw 'ByteString' (both public key and chain code)
 --
 -- @since 1.0.0
 xpubToBytes :: XPub -> ByteString
 xpubToBytes =
     CC.unXPub
 
--- | Convert an 'XPub' to a public key 'ByteString' without chain code included
+-- | Get a public key 'ByteString' from 'XPub'
 --
 -- @since 1.0.1
-xpubToBytesWithoutChainCode :: XPub -> ByteString
-xpubToBytesWithoutChainCode (CC.XPub pk _cc) =
+getPublicKey :: XPub -> ByteString
+getPublicKey (CC.XPub pk _cc) =
     pk
+
+-- | Get a chain code 'ByteString' from 'XPub'
+--
+-- @since 1.0.1
+getChainCode :: XPub -> ByteString
+getChainCode (CC.XPub _pk (CC.ChainCode cc)) =
+    cc
 
 -- | Construct an 'XPrv' from raw 'ByteString'.
 --
--- a/src/Cardano/Address/Derivation.hs
+++ b/src/Cardano/Address/Derivation.hs
@@ -31,6 +31,7 @@ module Cardano.Address.Derivation
     , XPub
     , xpubFromBytes
     , xpubToBytes
+    , xpubToBytesWithoutChainCode
 
     -- ** XPrv
     , XPrv
@@ -116,12 +117,19 @@ xpubFromBytes :: ByteString -> Maybe XPub
 xpubFromBytes =
     eitherToMaybe . CC.xpub
 
--- | Convert an 'XPub' to a raw 'ByteString'
+-- | Convert an 'XPub' to a raw 'ByteString' (public key plus chain code)
 --
 -- @since 1.0.0
 xpubToBytes :: XPub -> ByteString
 xpubToBytes =
     CC.unXPub
+
+-- | Convert an 'XPub' to a public key 'ByteString' without chain code included
+--
+-- @since 1.0.1
+xpubToBytesWithoutChainCode :: XPub -> ByteString
+xpubToBytesWithoutChainCode (CC.XPub pk _cc) =
+    pk
 
 -- | Construct an 'XPrv' from raw 'ByteString'.
 --
--- a/src/Cardano/Address/Style/Shelley.hs
+++ b/src/Cardano/Address/Style/Shelley.hs
@@ -25,6 +25,10 @@ module Cardano.Address.Style.Shelley
       -- * Staking
     , deriveStakingPrivateKey
 
+      -- * Discrimination
+    , MkNetworkDiscriminantError (..)
+    , mkNetworkDiscriminant
+
       -- * Accessors
     , getKey
 
@@ -248,6 +252,21 @@ instance HasNetworkDiscriminant Shelley where
     type NetworkDiscriminant Shelley = NetworkTag
     addressDiscrimination _ = RequiresNetworkTag
     networkTag = id
+
+-- | Error reported from trying to create a network discriminant from number
+--
+-- @since 1.0.0
+newtype MkNetworkDiscriminantError
+    = ErrWrongNetworkTag Word8
+      -- ^ Wrong network tag.
+    deriving (Eq, Show)
+
+mkNetworkDiscriminant
+    :: Word8
+    -> Either MkNetworkDiscriminantError (NetworkDiscriminant Shelley)
+mkNetworkDiscriminant nTag
+    | nTag < 16 =  Right $ NetworkTag $ fromIntegral nTag
+    | otherwise = Left $ ErrWrongNetworkTag nTag
 
 instance PaymentAddress Shelley where
     paymentAddress discrimination k = unsafeMkAddress $
--- a/src/Cardano/Address/Style/Shelley.hs
+++ b/src/Cardano/Address/Style/Shelley.hs
@@ -55,7 +55,7 @@ import Cardano.Address.Derivation
     , deriveXPrv
     , deriveXPub
     , generateNew
-    , xpubToBytesWithoutChainCode
+    , getPublicKey
     )
 import Cardano.Mnemonic
     ( someMnemonicToBytes )
@@ -66,7 +66,9 @@ import Control.Exception.Base
 import Crypto.Hash
     ( hash )
 import Crypto.Hash.Algorithms
-    ( Blake2b_224 )
+    ( Blake2b_224 (..) )
+import Crypto.Hash.IO
+    ( HashAlgorithm (hashDigestSize) )
 import Data.Binary.Put
     ( putByteString, putWord8, runPut )
 import Data.ByteArray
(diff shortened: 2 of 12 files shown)


src/Cardano/Address/Style/Shelley.hs
```diff
@@ -73,6 +92,35 @@ import qualified Data.ByteArray as BA
...
+-- > import Cardano.Address.Derivation ( AccountingStyle(..), GenMasterKey(..), toXPub )
+-- >
+-- > let accIx = toEnum 0x80000000
+-- > let acctK = deriveAccountPrivateKey rootK accIx
+-- >
+-- > let addIx = toEnum 0x00000014
+-- > let addrK = deriveAddressPrivateKey acctK UTxOExternal addIx
+-- >
+-- > bech32 $ paymentAddress mainnetDiscriminant (toXPub <$> addrK)
+-- > "addr1vzpfffuj3zkp5g7ct6h4va89caxx9ayq2gvkyfvww48sdncxzgg5v"
```
paweljakubas, line 123: something to investigate  : in cddl we have line like :
```
; 0110: enterprise address: keyhash28 -- 29 bytes -- 47 bech32
```
but I have :
```
"addr1vzpfffuj3zkp5g7ct6h4va89caxx9ayq2gvkyfvww48sdncxzgg5v
```
ie.. `addr` plus `54-characters`  after bech32

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -169,21 +216,46 @@ instance SoftDerivation Shelley where
...
+          -- discriminant in CDDL spec is known.
+          firstByte = case discrimination of
+              RequiresNoMagic  -> 0x60
+              RequiresMagic (ProtocolMagic _pm) -> 0x61
+          hashedKey =
+              BA.convert $
+              hash @_ @Blake2b_224 $
+              xpubToBytesWithoutChainCode $
+              getKey k
+          expectedLength = 1 + publicKeyHashSize
```
paweljakubas, line 242: we are ok here for enterprise - 1 + 28 = 29 bytes

src/Cardano/Address/Style/Shelley.hs
KtorZ: I wouldn't worry too much about this. In particular, this comment doesn't tell anything about which prefix is used, and the prefix has an influence on the integrated checksum and the encoding itself. What matters is the length of the original data.

paweljakubas: after recalling bech32 standard it is probably ok, as we have (adding to `addr` + `54-character` rest): 
- addr
- 1 (separator)
- 47-character payload
- 6-character checksum

src/Cardano/Address/Derivation.hs
```diff
@@ -116,13 +117,20 @@ xpubFromBytes :: ByteString -> Maybe XPub
...
 xpubToBytes :: XPub -> ByteString
 xpubToBytes =
     CC.unXPub
 
+-- | Convert an 'XPub' to a public key 'ByteString' without chain code included
+--
+-- @since 1.0.1
+xpubToBytesWithoutChainCode :: XPub -> ByteString
+xpubToBytesWithoutChainCode (CC.XPub pk _cc) =
+    pk
```
KtorZ, line 132: What about exposing `getPublicKey` and `getChainCode` instead ? This name doesn't look quite right :no_mouth:

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -169,21 +218,68 @@ instance SoftDerivation Shelley where
...
+invariantSize expectedLength bytes
+    | BS.length bytes == expectedLength = bytes
+    | otherwise = error
+      $ "length was "
+      ++ show (BS.length bytes)
+      ++ ", but expected to be "
+      ++ (show expectedLength)
+
+keyToHash :: Shelley depth XPub -> ByteString
+keyToHash = BA.convert .
```
KtorZ, line 271: I'd suggest calling it `blake2b224`, it reads better at call-sites.

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -169,21 +218,68 @@ instance SoftDerivation Shelley where
...
+    hash @_ @Blake2b_224 .
+    xpubToBytesWithoutChainCode .
```
KtorZ, line 273: Are we sure that we need to hash the public key only; or the extended public key with chain code ?

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -169,21 +218,68 @@ instance SoftDerivation Shelley where
...
+          -- rationale why we may need such addresses refer to delegation
+          -- specification - Section 3.2.3. What is important here is that the
+          -- address is composed of discrimination byte and 28 bytes hashed public key.
+          -- Moreover, it was decided that first 4 bits for enterprise address
+          -- will be `0110`. The next for bits for network discriminator are to
+          -- be established soon. TO_DO : change `firstbyte` as network
+          -- discriminant in CDDL spec is known.
+          firstByte = case discrimination of
+              RequiresNoMagic  -> 0x60
+              RequiresMagic (ProtocolMagic _pm) -> 0x61
```
KtorZ, line 238: We need to figure out how the network discrimination exactly works in Shelley. I am not sure that this byron `RequiresNoMagic` / `RequiresMagic` applies anymore. As far as I understand from the CDDL spec, it's now just a number, so I assume it's always present.

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -169,21 +218,68 @@ instance SoftDerivation Shelley where
...
--- Serialized length in bytes of a Single Address
-addrSingleSize :: Int
-addrSingleSize = 1 + publicKeySize
-
--- Serialized length in bytes of a Grouped Address
-addrGroupedSize :: Int
-addrGroupedSize = addrSingleSize + publicKeySize
+-- | Size, in bytes, of a hash of public key (without chain code)
+publicKeyHashSize :: Int
+publicKeyHashSize = 28
```
KtorZ, line 282: Note that instead of hard-coding the constant here, we could simply use: https://hackage.haskell.org/package/cryptonite-0.26/docs/Crypto-Hash-IO.html#v:hashDigestSize

```hs
hashDigestSize Blake2b_224
28 
```

test/Cardano/Address/Style/ShelleySpec.hs
```diff
@@ -62,7 +64,8 @@ prop_publicChildKeyDerivation
 prop_publicChildKeyDerivation (mw, sndFactor) cc ix =
     addrXPub1 === addrXPub2
   where
-    rootXPrv = genMasterKeyFromMnemonic mw sndFactor :: Shelley 'RootK XPrv
+    rootXPrv =
+        genMasterKeyFromMnemonic mw (maybeMnemonicToBytes sndFactor) :: Shelley 'RootK XPrv
```
KtorZ, line 68: What about changing the argument of the property instead to be `ScrubbedBytes` instead of `Maybe SomeMnemonic` :) ?

src/Cardano/Address/Style/Shelley.hs
paweljakubas: yes, as it was answered yesterday by Duncan and ledger team

test/Cardano/Address/Style/ShelleySpec.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/20/commits/b97b25b7bbf9c5f2130b016a8e88457325063549

src/Cardano/Address/Style/Shelley.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/20/commits/b97b25b7bbf9c5f2130b016a8e88457325063549

src/Cardano/Address/Style/Shelley.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/20/commits/b97b25b7bbf9c5f2130b016a8e88457325063549

src/Cardano/Address/Derivation.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/20/commits/b97b25b7bbf9c5f2130b016a8e88457325063549

src/Cardano/Address/Style/Shelley.hs
paweljakubas: So,
```
data NetworkDiscriminantByron = RequiresNoMagic | RequiresMagic ProtocolMagic
data NetworkDiscriminantShelley = NetworkId Word8
data NetworkDiscriminant = NetworkDiscriminantByron | NetworkDiscriminantShelley
```
?

paweljakubas: attempted here https://github.com/input-output-hk/cardano-addresses/pull/20/commits/0a200b8867ef0b792340d4fbfce7731ba87450f7

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -169,21 +222,74 @@ instance SoftDerivation Shelley where
...
+          -- we use here the fact that payment address stands for what is named
+          -- as enterprise address, ie., address carrying no stake rights. For
+          -- rationale why we may need such addresses refer to delegation
+          -- specification - Section 3.2.3. What is important here is that the
+          -- address is composed of discrimination byte and 28 bytes hashed public key.
+          -- Moreover, it was decided that first 4 bits for enterprise address
+          -- will be `0110`. The next for 4 bits are reserved for network discriminator.
+          -- `0110 0000` is 96 in decimal.
+          firstByte =
+              96 + invariantNetworkTag (networkTag @Shelley discrimination)
```
KtorZ, line 245: It'd be good to get a smart-constructor for `NetworkDiscriminant Shelley` that actually enforces this invariant. So that we can have: `mkNetworkDiscriminant :: Word8 -> Either XXX (NetworkDiscriminant Shelley)`. What do you think ?

test/Cardano/Address/Style/ShelleySpec.hs
```diff
@@ -85,3 +99,17 @@ prop_toEnumAccountingStyle n =
...
+
+newtype SndFactor = SndFactor ScrubbedBytes
+    deriving stock (Eq, Show)
+    deriving newtype (ByteArrayAccess)
+
+instance Arbitrary SndFactor  where
+    arbitrary = do
+        n <- choose (0, 64)
+        bytes <- BS.pack <$> vector n
+        return $ SndFactor $ BA.convert bytes
```
KtorZ, line 115: :+1:

src/Cardano/Address/Style/Shelley.hs
paweljakubas: indeed, done in https://github.com/input-output-hk/cardano-addresses/pull/20/commits/78dd471eda7cf18172ea19b127165f8a2896c146

PR #21, opened 2020-04-30
distinguish between staking key and address key

KtorZ:
This makes the API a bit clearer about what key is what. In the end, users needs not to know that the staking key is located at the same path in the derivation tree.

### DIFF
--- a/src/Cardano/Address/Style/Shelley.hs
+++ b/src/Cardano/Address/Style/Shelley.hs
@@ -22,6 +22,9 @@ module Cardano.Address.Style.Shelley
       -- * Shelley
       Shelley
 
+      -- * Staking
+    , deriveStakingPrivateKey
+
       -- * Accessors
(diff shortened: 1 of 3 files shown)


src/Cardano/Address/Style/Shelley.hs
```diff
@@ -249,7 +271,8 @@ instance DelegationAddress Shelley where
     delegationAddress discrimination paymentKey stakingKey = unsafeMkAddress $
         invariantSize expectedLength $ BL.toStrict $ runPut $ do
             putWord8 firstByte
-            mapM_ (putByteString . blake2b224) [paymentKey, stakingKey]
+            putByteString . blake2b224 $ paymentKey
+            putByteString . blake2b224 $ stakingKey
```
KtorZ, line 275: Can't `mapM_` anymore because they have different types.

PR #22, opened 2020-04-30
add Buildable derivation to mnemonic error

paweljakubas:
need this to port fully in cardano-wallet


paweljakubas:
In the end, after second thought, I added String to Buildable transformation at call site. So we will not need that

PR #23, opened 2020-05-04
correct code snippet and improve on documentation

paweljakubas:
1. Revisited ghci usage
2. Added missing docs for Shelley

### DIFF
--- a/src/Cardano/Address/Style/Shelley.hs
+++ b/src/Cardano/Address/Style/Shelley.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DerivingStrategies #-}
 {-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE InstanceSigs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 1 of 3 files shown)


src/Cardano/Address/Style/Shelley.hs
```diff
@@ -255,12 +261,16 @@ instance HasNetworkDiscriminant Shelley where
...
+-- @since 1.0.1
 newtype MkNetworkDiscriminantError
     = ErrWrongNetworkTag Word8
       -- ^ Wrong network tag.
     deriving (Eq, Show)
 
+-- | Construct 'NetworkDiscriminant' for Cardano 'Shelley' from number.
+-- If number is invalid `MkNetworkDiscriminantError
+--
+-- @since 1.0.1
```
KtorZ, line 273: Would be 2.0.0 I think with the changes we have done regarding the network discrimination.

src/Cardano/Address/Style/Shelley.hs
```diff
@@ -255,12 +261,16 @@ instance HasNetworkDiscriminant Shelley where
...
 mkNetworkDiscriminant
     :: Word8
     -> Either MkNetworkDiscriminantError (NetworkDiscriminant Shelley)
```
KtorZ, line 276: We need to make sure the documentation mentions the invariant and tells users that we expect a number between 0 and 15. Otherwise, it's kinda cryptic.

src/Cardano/Address/Style/Shelley.hs
paweljakubas: done

src/Cardano/Address/Style/Shelley.hs
paweljakubas: done

Issue #24, opened 2020-05-04
Port Jormungandr style


paweljakubas:
# Context

I need this porting to use cleanly cardano-addresses in cardano-wallet

# Decision

Port Jormungandr address derivation and corresponding testing here

# Acceptance Criteria

1. We have `Jormungandr` style added, with `GenMasterKey`, `HardDerivation`, `SoftDerivation`, `PaymentAddress` and `DelegationAddress` implemented.
2. We have corresponding tests added.
3. Proper documentation is added with examples of usage 

---

# Development 

# QA

PR #25, opened 2020-05-04
Port jormungandr style

paweljakubas:
Task:
#24 

Done:
1. Ported all needed interfaces from cardano-wallet
2. Added deriveStakingPrivatekey to DelegationAddress interface
3. Added unit tests

### DIFF
--- a/src/Cardano/Address.hs
+++ b/src/Cardano/Address.hs
@@ -28,6 +28,7 @@ module Cardano.Address
     , AddressDiscrimination (..)
     , NetworkTag (..)
     , invariantSize
+    , invariantNetworkTag
     ) where
 
 import Prelude
@@ -47,7 +48,7 @@ import Data.ByteString.Base58
 import Data.Text
     ( Text )
 import Data.Word
-    ( Word32 )
+    ( Word32, Word8 )
 import GHC.Generics
     ( Generic )
 import GHC.Stack
@@ -165,3 +166,12 @@ invariantSize expectedLength bytes
       ++ show (BS.length bytes)
       ++ ", but expected to be "
       ++ (show expectedLength)
+
+invariantNetworkTag :: HasCallStack => Word32 -> NetworkTag -> Word8
+invariantNetworkTag limit (NetworkTag num)
+    | num < limit = fromIntegral num
+    | otherwise = error
+      $ "network tag was "
+      ++ show num
+      ++ ", but expected to be less than "
+      ++ show limit
--- a/src/Cardano/Address.hs
+++ b/src/Cardano/Address.hs
@@ -33,7 +33,7 @@ module Cardano.Address
 import Prelude
 
 import Cardano.Address.Derivation
-    ( Depth (..), XPrv, XPub )
+    ( Depth (..), XPub )
 import Cardano.Codec.Cbor
     ( decodeAddress, deserialiseCbor )
 import Control.DeepSeq
@@ -117,7 +117,7 @@ class HasNetworkDiscriminant key => PaymentAddress key where
 
 -- | Encoding of delegation addresses for certain key types and backend targets.
 --
--- @since 1.0.1
+-- @since 2.0.0
 class PaymentAddress key
     => DelegationAddress key where
     -- | Convert a public key and a staking key to a delegation 'Address' valid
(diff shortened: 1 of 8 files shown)


src/Cardano/Address.hs
```diff
@@ -130,6 +134,19 @@ class PaymentAddress key
...
+    --
+    -- Deriving staking keys for something else than the initial account is not
+    -- recommended and can lead to incompatibility with existing wallet softwares
+    -- (Daedalus, Yoroi, Adalite...).
+    --
+    -- @since 2.0.0
+    deriveStakingPrivateKey
+        :: key 'AccountK XPrv
+        -> key 'StakingK XPrv
+
```
KtorZ, line 149: I don't think this is the right place to have this one. I'd suggest creating another type-class to capture this constraint as

```hs
class StakingDerivation key where
   ...
```

Very much like we have `SoftDerivation` and `HardDerivation`.

src/Cardano/Address/Style/Jormungandr.hs
```diff
@@ -0,0 +1,314 @@
...
+-- https://github.com/input-output-hk/implementation-decisions/blob/e2d1bed5e617f0907bc5e12cf1c3f3302a4a7c42/text/1852-hd-chimeric.md
+--
+-- It indicates that the subtree of this node is used according to this
+-- specification.
+--
+-- Hardened derivation is used at this level.
+purposeIndex :: Word32
+purposeIndex = 0x8000073c
+
+-- | One master node (seed) can be used for unlimited number of independent
```
KtorZ, line 296: Note that we don't want all these constants to show up in the documentation. Haddock will automatically prune these **if and only if** they have no docstring attached. So here, we need to replace `-- | ...` with `-- ...` for everything we don't want to be exposed.

src/Cardano/Address/Style/Jormungandr.hs
```diff
@@ -0,0 +1,314 @@
...
+instance DelegationAddress Jormungandr where
+    delegationAddress discrimination paymentKey stakingKey = unsafeMkAddress $
+        invariantSize expectedLength $ BL.toStrict $ runPut $ do
+            putWord8 firstByte
+            putByteString . getPublicKey . getKey $ paymentKey
+            putByteString . getPublicKey . getKey $ stakingKey
+      where
+          firstByte = case addressDiscrimination @Jormungandr discrimination of
+              RequiresNetworkTag -> 0x84
+              RequiresNoTag -> 0x04
```
KtorZ, line 250: These constants should be the value behind `networkTag @Jormungandr`. So here, I'd suggest to simply export a single smart constructor as:

```hs
incentivizedTestnet :: NetworkDiscriminat Jormungandr
incentivizedTestnet = (RequiresNetworkTag, 0x84)
``` 

And, simply use `networkTag @Jormungandr` for the first byte. The purpose of the `HasNetworkDiscriminant` abstraction is to capture these constants and offer a consistant API to the caller.

src/Cardano/Address/Style/Jormungandr.hs
```diff
@@ -0,0 +1,314 @@
...
+jormungandrTestnet :: NetworkDiscriminant Jormungandr
+jormungandrTestnet = (RequiresNetworkTag, NetworkTag 0)
+
+{-------------------------------------------------------------------------------
+                                 Key generation
+-------------------------------------------------------------------------------}
+
+-- | Size, in bytes, of a public key (without chain code)
+publicKeySize :: Int
+publicKeySize = 32
```
KtorZ, line 274: Same as for Shelley, it'd be better to pull the constant from cryptonite here. https://hackage.haskell.org/package/cryptonite-0.26/docs/Crypto-PubKey-Ed25519.html#v:publicKeySize

src/Cardano/Address.hs
paweljakubas: done

src/Cardano/Address/Style/Jormungandr.hs
paweljakubas: done

src/Cardano/Address/Style/Jormungandr.hs
paweljakubas: done

src/Cardano/Address/Style/Jormungandr.hs
paweljakubas: done

src/Cardano/Address/Style/Jormungandr.hs
```diff
@@ -0,0 +1,313 @@
...
+            putWord8 (invariantNetworkTag 255 firstByte)
+            putByteString (getPublicKey $ getKey k)
+      where
+          firstByte = networkTag @Jormungandr discrimination
+          expectedLength = 1 + publicKeySize
+
+instance DelegationAddress Jormungandr where
+    delegationAddress discrimination paymentKey stakingKey = unsafeMkAddress $
+        invariantSize expectedLength $ BL.toStrict $ runPut $ do
+            putWord8 (invariantNetworkTag 255 $ NetworkTag $ firstByte + 1)
```
KtorZ, line 260: `+1` :thinking:  ?

paweljakubas: for paymentAddress singleAddress is used, for delegationAddress - groupedAddress : https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/src/Cardano/Wallet/Primitive/AddressDerivation/Shelley.hs#L343-L344 - if I understand well

KtorZ: :+1:

PR #26, opened 2020-05-05
re-export specialized function from Cardano.Address for better documentation

KtorZ:
Before, modules would look quite empty unless one noticed the instance declaration below the main type. Now, the documentation is a lot clearer and gives less cryptic signatures for the various interfaces. For example:

Screenshot from 2020-05-05 15-37-19
Screenshot from 2020-05-05 15-37-30


KtorZ:
Because I was waiting for your feedback first before doing any more work :D

PR #27, opened 2020-05-06
Add test vectors

paweljakubas:
Adding golden tests for a number of interesting situations

### DIFF
--- a/test/Cardano/Address/Style/ShelleySpec.hs
+++ b/test/Cardano/Address/Style/ShelleySpec.hs
@@ -21,7 +21,6 @@ import Cardano.Address
     , PaymentAddress (..)
     , bech32
     , bech32WithHrp
-    , hex
     , unsafeMkAddress
     )
 import Cardano.Address.Derivation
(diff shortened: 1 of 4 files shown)


test/Cardano/Address/Style/ShelleySpec.hs
```diff
@@ -64,6 +84,618 @@ spec = do
     describe "Enum Roundtrip" $ do
         it "AccountingStyle" (property prop_roundtripEnumAccountingStyle)
 
+    describe "Golden tests" $ do
+        goldenTest TestVector
+            { rootXPrv =
+                    "root_xprv982a5cfec6be5656be247f0060d13d3bd52403bcca627f1d9\
```
KtorZ, line 90: This isn't a valid bech32 encoding :no_mouth: ... looks like an hex-encoded string to which a `root_xprv` prefix was appended.

paweljakubas: used bech32 in root prv key rather than hex -> https://github.com/input-output-hk/cardano-addresses/pull/27/commits/291413594b67e58e0c04fa4c714179be773a9b2f

PR #29, opened 2020-05-15
command for key derivation and recovery-phrase generation

KtorZ:
- 05fc02f0484f49bb3f2668700c7f514abc4068c7
  :round_pushpin: **start porting commands from cardano-wallet key & mnemonic**
  
- f1bcbfc2a3e7ddf29d92f4f46ac29cbaa4900e3d
  :round_pushpin: **rename 'mnemonic' command to 'recovery-phrase'**
  For better matching the current style guide & terminology company-wide

- 543dcfd57787e8d0a3b60e0ba2f0c3690d714caa
  :round_pushpin: **add support for 'key child' command**
  
- a5d0aee17fe098c80806cc8d8bb1d25d26c70c8d
  :round_pushpin: **implement key public and key inspect functions**
  
- 0c7ae6b2a199b07de69244171f1c7dc0a218e636
  :round_pushpin: **split command-line parsers & handlers into multiple organized modules**
  
- d2d7c0bf506f62865632f39cf6615046a5b31b7c
  :round_pushpin: **define test helper to easily run and collect output of the CLI**
  
- fcb5586b3b34b9d7cb23851ecdb6e6a34aed9ea5
  :round_pushpin: **port specs from cardano-wallet regarding mnemonic generation**
  
- 314889bebf8c0811dde271783b4b33d1042d9139
  :round_pushpin: **port and complete golden tests for root key generation**
  In the meantime, also improved the error message when failing to parse a style. This now provides
a better error message indicating what values are expected.

- fea5e936c8f1d632f8ac3e94358287d743635c63
  :round_pushpin: **port commute property between child <-> public for key derivation commands**
  
- c8a4de5b04c70e4c6f4db6a7b470af3a29d26369
  :round_pushpin: **complete coverage overlay for newly added functions**
  ignoring auto-generated instances from the coverage report

- b84c46679489663e2861ccd0f76897ff27fb946a
  :round_pushpin: **display command descriptions as spec description**
  
- 2c63a6966c33fe3abd2941af4155e1e7307655b3
  :round_pushpin: **add integration tests for remaining public / child / inspect sub-commands of 'key'**
  
- 580dde4bcbb3a7b919955980e59a677556cb2073
  :round_pushpin: **roundtrip and edge tests for Derivation applicative parsers**
  
- d16e18b1c72bb571fb1d759a9fbe4252a4a451da
  :round_pushpin: **add tests for System.IO.Extra, and review encoding detection in the meantime.**


src/System/IO/Extra.hs
```diff
@@ -0,0 +1,166 @@
...
+
+--
+-- I/O Read
+--
+
+-- | Read some bytes from the console, and decode them if the encoding is recognized.
+hGetBytes :: Handle -> IO ByteString
+hGetBytes h = do
+    raw <- B8.hGetContents h
+    case (tryBech32 raw <|> tryBase16 raw <|> tryBase58 raw) of
```
Anviking, line 73: No, I'm actually concerned here
```haskell
λ> tryAny "11"
"\DC1"
λ> tryBase58 "11"
Just "\NUL\NUL"
```
This function does not have access to the expected length of the decoded bytes, so it get's even trickier.

KtorZ: It doesn't actually matter much does it? The bytestrings manipulated by this program are at least 32 bytes (so ~64 characters minimum hex encoded).

There's no way a 32 bytestrings can be both base58 and base16 decoded.

Anviking: - What if you enter a `base58` key that is also valid hex-bytes (but not key), then wouldn't `hGetBytes` choose hex, and the key-from-bytes function fail, instead of successfully decoding a base58 key?
- I suspect error messages will be _much_ worse
- There is also (I believe) ambiguity of xpub (64) vs xprv (96). If someone in the future added something of size 128: we could silently miss that there is no way do distinguish between 128byte base58 and 64 byte hex.

Anviking: Some kind of top-level roundtrip tests (like in previous implementation) should catch these. (At least 1, and 3)

src/Command/Key/Child.hs
```diff
@@ -0,0 +1,88 @@
...
+mod :: (Cmd -> parent) -> Mod CommandFields parent
+mod liftCmd = command "child" $
+    info (helper <*> fmap liftCmd parser) $ mempty
+        <> progDesc "Derive child keys from a parent public/private key."
+        <> footerDoc (Just $ string $ mconcat
+            [ "The parent key is read from stdin."
+            ])
+  where
+    parser = Child
+        <$> encodingOpt [humanReadablePart|???|]
```
paweljakubas, line 58: what ??? exactly suppose to do here ?

KtorZ: The `child` command works for both `xprv` and `xpub`. We only know what sort of key is given when handling the command. Note also that the "encoding" is only use for the _output_, not for decoding inputs (here we don't assume any encoding, and we try to decode any known encoding). 

What happens is that, I simply `fmap` over the encoding when the right prefix is known, to replace the `???` with an appropriate value:

https://github.com/input-output-hk/cardano-addresses/pull/29/files#diff-63fa7f406d9975760ab8aa7a21b2260dR77
```hs
                    pure
                        ( xpubToBytes child
                        , fmap (const [humanReadablePart|xpub|] ) encoding
                        )
```

src/System/IO/Extra.hs
KtorZ: > What if you enter a base58 key that is also valid hex-bytes (but not key), then wouldn't hGetBytes choose hex, and the key-from-bytes function fail, instead of successfully decoding a base58 key?

What do you mean?

> I suspect error messages will be much worse

At this stage, they will be worse indeed. 

> There is also (I believe) ambiguity of xpub (64) vs xprv (96). If someone in the future added something of size 128: we could silently miss that there is no way do distinguish between 128byte base58 and 64 byte hex.

I disagree here. The alphabet is not the same! The probability that a string of 50 characters encoded in Base58 would only contain characters from the base16 alphabet is 15/58**50 ~ 1e-30 !! And I am not even sure that this makes for a valid encoding! Really, I don't think there's any issue.

Anviking: > What do you mean?

I was imagining "that a string in Base58 would [...] only contain characters from the base16"

> The probability that a string of

Ok, fair point.

Then just: You wrote "especially if base16 is tried last", but doesn't seem to be done here

Also ran in to something strange
```
# A root key
$ cardano-wallet-shelley key root -- art forum devote street sure rather head chuckle guard poverty release quote oak craft enemy
b8f2bece9bdfe2b0282f5bad705562ac996efb6af96b648f4445ec44f47ad95c10e3d72f26ed075422a36ed8585c745a0e1150bcceba2357d058636991f38a3791e248de509c070d812ab2fda57860ac876bc489192c1ef4ce253c197ee219a4

# Can derive child from recovery phrase using cardano-address 
$ echo "art forum devote street sure rather head chuckle guard poverty release quote oak craft enemy" | cardano-address key from-recovery-phrase icarus --base16 | cardano-address key child 0
xprv13zahs5rgyl8a60xkaxkpucvlz48fl08qtq2cukatde5nr7r6m9w0xnjkvvaq233mf045q7wn2h83t65f85l95au049kn3yk35vwla2w72hpcpm2v0e0ezkd6kh3p823qn2n3zmle200en5t3whxf5zj375kw8aes

# ... but can not go from the root key directly?
$ echo "b8f2bece9bdfe2b0282f5bad705562ac996efb6af96b648f4445ec44f47ad95c10e3d72f26ed075422a36ed8585c745a0e1150bcceba2357d058636991f38a3791e248de509c070d812ab2fda57860ac876bc489192c1ef4ce253c197ee219a4" | cardano-address key child "0"
Couldn't convert bytes into neither extended public or private keys.

# Works with old
$ echo "b8f2bece9bdfe2b0282f5bad705562ac996efb6af96b648f4445ec44f47ad95c10e3d72f26ed075422a36ed8585c745a0e1150bcceba2357d058636991f38a3791e248de509c070d812ab2fda57860ac876bc489192c1ef4ce253c197ee219a4" | cardano-wallet-shelley key child --path 0
88bb78506827cfdd3cd6e9ac1e619f154e9fbce058158e5bab6e6931f87ad95cf34e56633a05463b4beb4079d355cf15ea893d3e5a778fa96d3892d1a31dfea9de55c380ed4c7e5f9159bab5e213aa209aa7116ff953df99d17175cc9a0a51f5

```

KtorZ:
@paweljakubas @Anviking added tests, some from cardano-wallet, some specially crafted for this purpose. I caught a few small bugs along the way. 

I also added a 'Command-Line' section on the README with some key examples.

src/System/IO/Extra.hs
KtorZ:
```console
# ... but can not go from the root key directly?
$ echo "b8f2bece9bdfe2b0282f5bad705562ac996efb6af96b648f4445ec44f47ad95c10e3d72f26ed075422a36ed8585c745a0e1150bcceba2357d058636991f38a3791e248de509c070d812ab2fda57860ac876bc489192c1ef4ce253c197ee219a4" | cardano-address key child "0"
Couldn't convert bytes into neither extended public or private keys.
```

Indeed! `echo` adds a trailing `\n` that needs to be trimmed. Now fixed.

PR #32, opened 2020-06-01
Separate word7 module

paweljakubas:
Follow-up pointer addresses PR
- [x] separated Word7 in module
- [x] used it in other modules
- [x] added property tests for roundtrips

### DIFF
--- a/src/Cardano/Address/Style/Shelley.hs
+++ b/src/Cardano/Address/Style/Shelley.hs
@@ -98,7 +98,7 @@ import Data.Maybe
 import Data.Word
     ( Word32, Word8 )
 import Data.Word7
-    ( limit, putVariableLengthNat )
+    ( putVariableLengthNat )
 import GHC.Generics
     ( Generic )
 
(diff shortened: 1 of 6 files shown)


src/Cardano/Address/Style/Shelley.hs
```diff
@@ -88,21 +88,19 @@ import Crypto.Hash.Algorithms
...
     ( ScrubbedBytes )
 import Data.ByteString
     ( ByteString )
 import Data.Maybe
     ( fromMaybe )
 import Data.Word
-    ( Word32, Word64, Word8 )
+    ( Word32, Word8 )
+import Data.Word7
+    ( limit, putVariableLengthNat )
```
KtorZ, line 101: since `limit` is a rather ambiguous identifier, I'd very much suggest to use a qualified import for it. Like `filter` from `Data.Map.Strict` for instance. `Map.filter ...` provides the necessary context. `Word7.limit ...` would be less ambiguous.

paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/32/commits/c0cec9da20e038b1cdcc3e63a5953660a1784764

PR #35, opened 2020-06-16
Docker entrypoint fix and docs

rhyslbw:
- Fixes the Docker entrypoint that was using to the build step dir structure
- Adds documentation to the README for building and running commands using auto-remove containers
- Removes whitespace in README


KtorZ:
Thanks @rhyslbw :heart:

PR #38, opened 2020-06-17
`inspect` commands return structured JSON

KtorZ:
- 0c4e5bcce52c6d746db29aa3d3043eb1a38f65df
  :round_pushpin: **return structured JSON instead of pretty string for address introspection**
  So that the output can be easily parsed and used by another software.

TODO: `key inspect` as well for consistency.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -30,7 +30,7 @@
 
 - modules:
   # Enforce some common qualified imports aliases across the codebase
-  - {name: [Data.Aeson, Data.Aeson.Types], as: Aeson}
+  - {name: [Data.Aeson, Data.Aeson.Types], as: Json}
   - {name: [Data.ByteArray], as: BA}
   - {name: [Data.ByteString.Base16], as: B16}
   - {name: [Data.ByteString.Char8], as: B8}
--- a/README.md
+++ b/README.md
@@ -109,24 +109,37 @@ xpub16y4vhpyuj2t84gh2qfe3ydng3wc37yqzxev6gce380fvvg47ye8um3dm3wn5a64gt7l0fh5j6sj
 </details>
 
 ## Docker Image
+
 ### Build
-```
+
+```console
 $ docker build -t cardano-address .
 ```
+
 ### Run
+
 Use the auto-remove flag `--rm` when running commands.
-```
+
+```console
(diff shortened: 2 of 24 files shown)


rhyslbw:
Great work @KtorZ. I've defined types for Daedalus (Flow) as follows:

``` ts
export type IntrospectAddressRequest = {
  input: string
};

export type AddressStyle = 'Byron' | 'Icarus' | 'Jormungandr' | 'Shelley'

export type ChainPointer = {
  slot_num: number,
  transaction_index: number,
  output_index: number
}

export type AddressBase = {
  address_style: AddressStyle,
  network_tag: number | null
}

export type ByronAddress = AddressBase & {
  stake_reference: 'none',
  address_root: string,
  derivation_path: string,
}

export type IcarusAddress = AddressBase & {
  stake_reference: 'none',
  address_root: string,
}

export type JormungandrAddress = AddressBase & {
  address_type: 'single' |'group' | 'account' | 'multisig',
  account_key?: string,
  merkle_root?: string,
  stake_reference: 'by value' | 'none',
  spending_key?: string,
  stake_key?: string,
}

export type ShelleyAddress = AddressBase & {
  pointer?: ChainPointer,
  stake_reference: 'by pointer' | 'none',
  script_hash?: string,
  spending_key_hash?: string,
  stake_key_hash?: string,
  stake_script_hash?: string,
}

export type IntrospectAddressResponse = {
  introspection: ByronAddress | IcarusAddress | JormungandrAddress | ShelleyAddress
} | 'Invalid';

```

Is `AddressBase.network_tag` typed correctly for all styles?

KtorZ:
@rhyslbw there are 3 fields that are on every address, so I'd suggest `AddressBase` as follows:

```ts
export type AddressBase = {
  address_style: AddressStyle,
  network_tag: number | null,
  stake_reference: 'none' | 'by value' | 'by pointer'
}
```

Also, the `network_tag` can only be `null` for `Icarus` & `Byron` addresses. It'll necessarily exists on Shelley & Jormungandr addresses. Not sure if this needs to be captured.

rhyslbw:
> Not sure if this needs to be captured.
If we're merging the two `stake_reference` types then I think this is fine for the current use-case

Issue #41, opened 2020-07-09
cardano-address.exe --help fails on windows


rvl:
On Windows 10 and Wine, `cardano-address --help` fails with the following error:

```
cardano-address.exe: <stdout>: hPutChar: invalid argument (invalid character)
```

![VirtualBox_Win10_09_07_2020_11_14_19](https://user-images.githubusercontent.com/1019641/86985901-ae055000-c1d5-11ea-8a89-9c8db3dcf2ed.png)

KtorZ:
Let's setup a Windows CI with github actions and see what the tests spit out. I believe we'll need to:

a) be careful about newlines and use platform-agnostic primitives instead of raw `\n`
b) make sure that UTF-8 characters are correctly output (we have some helper in cardano-wallet for this) 
c) pray.

KtorZ:
I started toying around here: https://github.com/KtorZ/cardano-addresses/runs/854593414?check_suite_focus=true

And indeed, I can see that tests failing with some interesting logs similar to the above.

rvl:
Yes you can't simply output utf-8 encoded text and hope that it will work.
It's possible to avoid the exception by forcing haskell to use utf-8 for its output handles.

But if the system encoding is not utf-8 then you will see mojibake.
On windows you can avoid that by changing the code page of the console.
On linux you can't do that, but most linux systems default to utf-8 anyway.

KtorZ:
>  On windows you can avoid that by changing the code page of the console.

That's what I did yesterday indeed but it didn't quite help in CI at least (I believe Github Windows servers are configured to use utf-8 in the first place, but it might help in some other cases). 

There are three tests failing at the moment on Windows:

```
test\Command\Address\PointerSpec.hs:46:5: 
malformed pointer 💩 💩 💩
       "cannot parse value `cardano-address: <stderr>: hPutChar: invalid argument (invalid character)\n" does not contain "Usage"
```

```
test\Command\Address\PaymentSpec.hs:41:5: 
malformed network 💩
       "option --network-tag: cannot parse value `cardano-address: <stderr>: hPutChar: invalid argument (invalid character)\n" does not contain "Usage"
```

```
test\Command\Address\BootstrapSpec.hs:57:5: 
invalid network 💩
       "option --network-tag: cannot parse value `cardano-address: <stderr>: hPutChar: invalid argument (invalid character)\n" does not contain "Usage"
```

I guess the "pattern" is quite obvious.

[...]

PR #43, opened 2020-07-10
Force input/output text encoding to be utf-8

rvl:
Related to #41.

I didn't test anything yet but the same code works in cardano-wallet.

Depending on the test cases, it might also need `hSetEncoding utf8 stdout` in `test/Main.hs` so that tests pass (annoying to do with hspec-discover).

### DIFF
--- a/src/Command.hs
+++ b/src/Command.hs
@@ -34,7 +34,15 @@ import Options.Applicative.Help.Pretty
 import System.Console.ANSI
     ( hSupportsANSIWithoutEmulation )
 import System.IO
-    ( BufferMode (..), Handle, hSetBuffering, stderr, stdout )
+    ( BufferMode (..)
+    , Handle
+    , hSetBuffering
+    , hSetEncoding
+    , mkTextEncoding
+    , stderr
+    , stdin
+    , stdout
+    )
 import System.IO.Extra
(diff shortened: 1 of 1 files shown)


KtorZ:
@rvl I've been doing some experiments on my personal fork https://github.com/KtorZ/cardano-addresses/commits/master where I setup a Windows workflow as well.

I went for a fork at first as I didn't wanted to "clutter" the source repository with all my attempts to setup the Windows workflow but it went pretty smoothly in the end. I've added pretty much the same thing in there, yet, there are still 3 failing tests.

rvl:
OK - I suspect it's because the test suite also needs the text encoding set to utf-8.
Replace `test/Main.hs` with something like this:
https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Main.hs

KtorZ:
> OK - I suspect it's because the test suite also needs the text encoding set to utf-8.

That is indeed possible, although the test runners outputs the "problematic" characters just fine already in the test description  :shrug: 

> Replace test/Main.hs with something like this:
https://github.com/input-output-hk/cardano-wallet/blob/master/lib/core/test/unit/Main.hs

I'll try this out!

PR #44, opened 2020-07-10
Windows Issues / Testing

KtorZ:
- b7ac636ea8738fe86e0836f90f946efefe6b1cf3
  :round_pushpin: **use pre-define 'hashFiles' function for cache**
  
- 9d633beaea9f63a486ac8c751ebf35fc622349db
  :round_pushpin: **add workflow for Windows**
  
- d824f16fb34edb00cec24adf65091312821f38f4
  :round_pushpin: **force UTF-8 encoding, with Windows tricks**
  
- 3df6364cec050ee1ce7c5f2e3432952b18bb3637
  :round_pushpin: **do not spit back malformed network tag when given gibberish**
  It's usually not a good idea to spit back to users something that we failed to parse in the first place. Instead, we should print what we would have expected.

- 598c7c8536d79e7cb0b3fd93919ed925a0490347
  :round_pushpin: **remove unused pragma (new HLint is better at catching those)**

### DIFF
--- a/src/Command.hs
+++ b/src/Command.hs
@@ -10,6 +10,9 @@ module Command
     , setup
     , parse
     , run
+
+    -- * I/O Fix
+    , withUtf8Encoding
     ) where
 
 import Prelude
@@ -86,7 +89,7 @@ cli = info (helper <*> parser) $ mempty
 
 -- | Run a given command
 run :: CLI -> IO ()
-run = withUtf8Encoding . handle prettyIOException . \case
+run = handle prettyIOException . \case
     RecoveryPhrase sub -> RecoveryPhrase.run sub
     Key sub -> Key.run sub
     Address sub -> Address.run sub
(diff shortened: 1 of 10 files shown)


KtorZ:
I am now down to 1 error on windows

```
 Finished in 39.8719 seconds
187 examples, 1 failure
```

```
  test\Command\Address\PointerSpec.hs:46:5: 
  1) Command.Address.Pointer, Create addresses with a pointer that indicate the position of a registered stake
address on the chain.

, malformed pointer 💩 💩 💩
       "cannot parse value `cardano-address: <stderr>: hPutChar: invalid argument (invalid character)\n" does not contain "Usage"
```

src/Command.hs
```diff
@@ -76,7 +86,7 @@ cli = info (helper <*> parser) $ mempty
 
 -- | Run a given command
 run :: CLI -> IO ()
-run = handle prettyIOException . \case
+run = withUtf8Encoding . handle prettyIOException . \case
```
rvl, line 89: I read the test case which failed... the text decode failure happens during option parsing.
This is setting the encoding too late. The option parsing is already done.

KtorZ: Good point.

Issue #46, opened 2020-07-15
Can't inspect stake_test... addresses


adascanBovich:
Staking address created using commands:

`cardano-cli shelley stake-address key-gen --verification-key-file staking.vkey --signing-key-file staking.skey`

`cardano-cli shelley stake-address build --staking-verification-key-file staking.vkey > staking.addr`

`shelley@localhost:~/keys/keys1$ cat staking.addr`

`stake_test1uqr6kpjry04lu9v2qu989r4vtdks202e8fz70z8mgzj96ugtnm6ld`

Now trying to use _inspect_ command for it:

<img width="731" alt="Screenshot 2020-07-15 at 8 08 30" src="https://user-images.githubusercontent.com/58727428/87505790-6e75b280-c672-11ea-8f8a-216275f68693.png">

Make sure that _inspect_ command is working for addr_test.. address:

<img width="731" alt="Screenshot 2020-07-15 at 8 08 47" src="https://user-images.githubusercontent.com/58727428/87505871-a2e96e80-c672-11ea-91a7-929574ee7e75.png">

hasufell:
## QA

Changeset: https://github.com/input-output-hk/cardano-addresses/commit/9a22296d40fe7df30e9258ce151076e792b15d12

The primary change that meets the acceptance criteria is:

https://github.com/input-output-hk/cardano-addresses/blob/9a22296d40fe7df30e9258ce151076e792b15d12/core/lib/Cardano/Address/Style/Shelley.hs#L511-L526

This corresponds to the ledger spec:

https://github.com/input-output-hk/cardano-ledger-specs/blob/736c9295f5896de28fbcedc6a8fc4d509fb78d10/shelley/chain-and-ledger/shelley-spec-ledger-test/cddl-files/shelley.cddl#L98-L99

The tests are at:

https://github.com/input-output-hk/cardano-addresses/blob/9a22296d40fe7df30e9258ce151076e792b15d12/command-line/test/Command/Address/InspectSpec.hs#L42-L48

These tests/hashes must fail with an older version of `cardano-addresses` (without the code above).

piotr-iohk:
lgtm. I thought it might be good to add additional tests -> https://github.com/input-output-hk/cardano-addresses/pull/61

PR #56, opened 2020-08-17
Allow to inspect reward accounts wrt #46

hasufell:
Run:
- [x] tests
- [x] stylish
- [x] hlint

Built:
- [x] cabal
- [x] stack
- [x] nix

### DIFF
--- a/command-line/test/Command/Address/InspectSpec.hs
+++ b/command-line/test/Command/Address/InspectSpec.hs
@@ -39,6 +39,14 @@ spec = describeCmd [ "address", "inspect" ] $ do
     specInspectAddress ["Shelley", "by pointer"]
         "addr1gw2fxv2umyhttkxyxp8x0dlpdt3k6cwng5pxj3jhsydzer5ph3wczvf2x4v58t"
 
+    -- reward account: keyhash28
+    specInspectAddress ["Shelley", "by value", "stake_key_hash"]
+        "stake1upshvetj09hxjcm9v9jxgunjv4ehxmr0d3hkcmmvdakx7mqcjv83c"
+
+    -- reward account: scripthash32
+    specInspectAddress ["Shelley", "by value", "script_hash"]
+        "stake17pshvetj09hxjcm9v9jxgunjv4ehxmr0d3hkcmmvdakx7mrgdp5xscfm7jc"
+
     specInspectMalformed
         "💩"
 
--- a/core/lib/Cardano/Address/Style/Shelley.hs
+++ b/core/lib/Cardano/Address/Style/Shelley.hs
@@ -55,6 +55,7 @@ module Cardano.Address.Style.Shelley
       -- Internals
     , minSeedLengthBytes
     , pubkeyHashSize
+    , scriptHashSize
     ) where
 
 import Prelude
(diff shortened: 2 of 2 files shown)


core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -507,6 +507,23 @@ inspectShelleyAddress addr
...
+                        , "stake_key_hash"    .= base16 (BS.take pubkeyHashSize rest)
+                        , "network_tag"       .= network
+                        ]
+               -- 1111: reward account: scripthash28
+                0b11110000 | BS.length rest == scriptHashSize ->
+                    Just $ Json.object
+                        [ "address_style"     .= Json.String "Shelley"
+                        , "stake_reference"   .= Json.String "by value"
+                        , "stake_key_hash"    .= base16 (BS.take scriptHashSize rest)
+                        , "script_hash"       .= base16 (BS.take scriptHashSize rest)
```
KtorZ, line 524: In the case of stake address, there is either a script hash or a public stake key hash, there can't be both.

hasufell: I'm still struggling to understand what those all are, even after reading the design spec :hot_face:

hasufell:
> The PR should come with additional tests cases covering this two new cases. See existing scenarios in: https://github.com/input-output-hk/cardano-addresses/blob/master/command-line/test/Command/Address/InspectSpec.hs#L17

done

hasufell:
To consider: allowing to generate these addresses via cardano-addresses.

command-line/test/Command/Address/InspectSpec.hs
```diff
@@ -39,6 +39,14 @@ spec = describeCmd [ "address", "inspect" ] $ do
     specInspectAddress ["Shelley", "by pointer"]
         "addr1gw2fxv2umyhttkxyxp8x0dlpdt3k6cwng5pxj3jhsydzer5ph3wczvf2x4v58t"
 
+    -- reward account: keyhash28
+    specInspectAddress ["Shelley", "by value"]
+        "addr1upshvetj09hxjcm9v9jxgunjv4ehxmr0d3hkcmmvdakx7mq726p32"
+
+    -- reward account: scripthash32
+    specInspectAddress ["Shelley", "by value"]
+        "addr17pshvetj09hxjcm9v9jxgunjv4ehxmr0d3hkcmmvdakx7mrgdp5xsnxe5e6"
```
KtorZ, line 48: bech32 encoding is only a presentation layer so this works just fine. Yet in principle, stake addresses should use the prefix `stake` to distinguish them from standard addresses. We have started to work on a list of agreed prefixes to facilitate interop between the various tools we have, so have a quick look at: https://github.com/cardano-foundation/CIPs/tree/master/CIP5

command-line/test/Command/Address/InspectSpec.hs
```diff
@@ -39,6 +39,14 @@ spec = describeCmd [ "address", "inspect" ] $ do
     specInspectAddress ["Shelley", "by pointer"]
         "addr1gw2fxv2umyhttkxyxp8x0dlpdt3k6cwng5pxj3jhsydzer5ph3wczvf2x4v58t"
 
+    -- reward account: keyhash28
+    specInspectAddress ["Shelley", "by value"]
+        "addr1upshvetj09hxjcm9v9jxgunjv4ehxmr0d3hkcmmvdakx7mq726p32"
+
+    -- reward account: scripthash32
+    specInspectAddress ["Shelley", "by value"]
```
KtorZ, line 47: Note that you could make the command superfluous and the test stronger by requiring `"script_hash"` to be present in the output (and respectively, "stake_key_hash") for the above.

PR #57, opened 2020-08-17
Use MonadThrow for inspecting addresses

hasufell:
Wrt #46. Easier tracing of error causes.

Run:
- [x] tests
- [x] stylish
- [x] hlint

Built with:
- [x] cabal
- [x] stack
- [x] nix

### DIFF
--- a/command-line/lib/Command/Address/Inspect.hs
+++ b/command-line/lib/Command/Address/Inspect.hs
@@ -1,7 +1,11 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
 
 {-# OPTIONS_HADDOCK hide #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
 
 module Command.Address.Inspect
     ( Cmd (..)
@@ -13,7 +17,7 @@ import Prelude hiding
     ( mod )
 
 import Cardano.Address
-    ( unsafeMkAddress )
+    ( Address, unsafeMkAddress )
 import Cardano.Address.Style.Byron
     ( inspectByronAddress )
 import Cardano.Address.Style.Icarus
@@ -24,15 +28,24 @@ import Cardano.Address.Style.Shelley
     ( inspectShelleyAddress )
 import Control.Applicative
     ( (<|>) )
+import Control.Exception
+    ( SomeException, displayException )
+import Control.Monad.Error.Class
+    ( Error )
+import Fmt
+    ( format )
 import Options.Applicative
     ( CommandFields, Mod, command, footerDoc, helper, info, progDesc )
 import Options.Applicative.Help.Pretty
     ( bold, indent, string, vsep )
+import System.Exit
+    ( die )
 import System.IO
     ( stdin, stdout )
 import System.IO.Extra
     ( hGetBytes, progName )
 
+import qualified Data.Aeson as Json
 import qualified Data.Aeson.Encode.Pretty as Json
 import qualified Data.ByteString.Lazy.Char8 as BL8
 
@@ -64,15 +77,26 @@ mod liftCmd = command "inspect" $
   where
     parser = pure Inspect
 
+
+-- used for 'inspect'
+instance Error SomeException
+
 run :: Cmd -> IO ()
 run Inspect = do
     bytes <- hGetBytes stdin
     case inspect (unsafeMkAddress bytes) of
-        Nothing   -> fail "Unrecognized address on standard input"
-        Just json -> BL8.hPutStrLn stdout (Json.encodePretty json)
+      Right json -> BL8.hPutStrLn stdout (Json.encodePretty json)
+      Left  e    -> die $ format "Error: {}" (displayException e)
   where
-    inspect addr =
-        inspectByronAddress       addr <|>
-        inspectIcarusAddress      addr <|>
-        inspectJormungandrAddress addr <|>
-        inspectShelleyAddress     addr
+    -- We can't use IO here, because (<|>) in IO only catches IOException type,
+    -- but MonadThrow in IO doesn't specialize to IOException.
+    --
+    -- Luckily, there's an existing instance for:
+    --   instance (Error e) => Alternative (Either e)
+    inspect :: (e ~ SomeException) => Address -> Either e Json.Value
+    inspect addr = do
+      foldr1 (<|>)
+        [inspectByronAddress       addr
+        ,inspectIcarusAddress      addr
+        ,inspectJormungandrAddress addr
+        ,inspectShelleyAddress     addr]
--- a/core/lib/Cardano/Address/Errors.hs
+++ b/core/lib/Cardano/Address/Errors.hs
@@ -1,0 +1,93 @@
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+-- |
+-- Copyright: © 2020 IOHK
+-- License: Apache-2.0
+
+module Cardano.Address.Errors
+    ( -- $overview
+
+      -- * AddrError
+      ShelleyAddrError(..)
+    , ByronAddrError(..)
+    , IcarusAddrError(..)
+    , JormungandrAddrError(..)
+
+      -- * pretty printers
+    , prettyShError
+    , prettyBrError
+    , prettyIcError
+    , prettyJoError
+  ) where
+
+import Prelude
+
+import Control.Exception
+    ( Exception, displayException )
+import Fmt
+    ( format )
+
+
+-- $overview
+--
+-- This module provides error/exception types for Addresses.
+
+data ShelleyAddrError
+    = ShUnknownAddrType
+    | ShWrongInputSize Int -- the actual size
+    | ShPtrRetrieveError String -- human readable error of underlying operation
+
+deriving instance Show ShelleyAddrError
+
+instance Exception ShelleyAddrError where
+  displayException = prettyShError
+
+prettyShError :: ShelleyAddrError -> String
+prettyShError ShUnknownAddrType = "Unknown address type"
+prettyShError (ShWrongInputSize i) = format "Wrong input size of {}" i
+prettyShError (ShPtrRetrieveError s) = format "Failed to retrieve pointer (underlying errors was: {})" s
+
+
+data ByronAddrError
+    = BrMissingExpectedDerivationPath
+    | forall e . (Exception e, Show e) => BrDeserialiseError e
+
+deriving instance Show ByronAddrError
+
+instance Exception ByronAddrError where
+  displayException = prettyBrError
+
+prettyBrError :: ByronAddrError -> String
+prettyBrError BrMissingExpectedDerivationPath = "Missing expected derivation path"
+prettyBrError (BrDeserialiseError e) = format "Deserialisation error (was: {})" (show e)
+
+
+data IcarusAddrError
+    = IcUnexpectedDerivationPath
+    | forall e . (Exception e, Show e) => IcDeserialiseError e
+
+deriving instance Show IcarusAddrError
+
+instance Exception IcarusAddrError where
+  displayException = prettyIcError
+
+prettyIcError :: IcarusAddrError -> String
+prettyIcError IcUnexpectedDerivationPath = "Unexpected derivation path"
+prettyIcError (IcDeserialiseError e) = format "Deserialisation error (was: {})" (show e)
+
+
+data JormungandrAddrError
+    = JoUnknownAddrType
+    | JoWrongInputSize Int -- the actual size
+
+deriving instance Show JormungandrAddrError
+
+instance Exception JormungandrAddrError where
+  displayException = prettyJoError
+
+prettyJoError :: JormungandrAddrError -> String
+prettyJoError JoUnknownAddrType = "Unknown address type"
+prettyJoError (JoWrongInputSize i) = format "Wrong input size of {}" i
+
--- a/core/lib/Cardano/Address/Style/Byron.hs
+++ b/core/lib/Cardano/Address/Style/Byron.hs
@@ -75,6 +75,8 @@ import Cardano.Address.Derivation
     , toXPub
     , xpubToBytes
     )
+import Cardano.Address.Errors
+    ( ByronAddrError (..) )
 import Cardano.Mnemonic
     ( SomeMnemonic (..), entropyToBytes, mnemonicToEntropy )
 import Codec.Binary.Encoding
@@ -83,6 +85,8 @@ import Control.DeepSeq
     ( NFData )
 import Control.Exception.Base
     ( assert )
+import Control.Monad.Catch
+    ( MonadThrow, throwM )
 import Crypto.Hash
     ( hash )
 import Crypto.Hash.Algorithms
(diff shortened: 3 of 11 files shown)


command-line/lib/Command/Address/Inspect.hs
```diff
@@ -67,10 +69,10 @@ mod liftCmd = command "inspect" $
 run :: Cmd -> IO ()
 run Inspect = do
     bytes <- hGetBytes stdin
-    case inspect (unsafeMkAddress bytes) of
-        Nothing   -> fail "Unrecognized address on standard input"
-        Just json -> BL8.hPutStrLn stdout (Json.encodePretty json)
+    json <- inspect (unsafeMkAddress bytes)
+    BL8.hPutStrLn stdout (Json.encodePretty json)
   where
+    inspect :: Address -> IO Json.Value
```
KtorZ, line 75: :+1:

core/lib/Cardano/Address/Errors.hs
```diff
@@ -0,0 +1,42 @@
...
+--
+-- This module provides error/exception types for Addresses.
+
+data AddrError = UnknownAddrType Address
+               | WrongInputSize Address
+                                Int -- the actual size
+               | PtrRetrieveError Address
+                                  String -- human readable error of underlying operation
+               | InvalidDerivationPath Address
+               | forall e . (Exception e, Show e) => DeserialiseError Address e
```
KtorZ, line 32: Avoid variable-length indentation whenever possible. It plays poorly with git diff. Consider instead:

```hs
data AddrError
    = UnknownAddrType Address
    | WrongInputSize Address Int 
    | PtrRetrieveError Address String
    | InvalidDerivationPath Address 
    | forall e. (Exception e, Show e) => DeserialiseError Address e
```

KtorZ: see also: https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#avoid-variable-length-indentation

core/lib/Cardano/Address/Errors.hs
```diff
@@ -0,0 +1,42 @@
...
+
+instance Show AddrError where
+  show (UnknownAddrType addr)       = "UnknownAddrType "       ++ show addr
+  show (WrongInputSize addr int)    = "WrongInputSize "        ++ show addr ++ " " ++ show int
+  show (PtrRetrieveError addr str)  = "PtrRetrieveError "      ++ show addr ++ " " ++ str
+  show (InvalidDerivationPath addr) = "InvalidDerivationPath " ++ show addr
+  show (DeserialiseError addr e)    = "DeserialiseError "      ++ show addr ++ " " ++ show e
```
KtorZ, line 39: It is generally **not** a good idea to manually define custom stock instances like `Show`, `Read` or `Ord`. Instead, consider privileging another type-class that is meant for displaying human-readable text input (Show really is made for debugging). 

More importantly, the malformed addresses should not be part of the output string. This can be an arbitrarily large / damaging payload which we certainly don't want to spit back to users. Instead, the error should gives examples of correct values whenever possible, or simply give hint about the failure in the input data.

core/lib/Cardano/Address/Errors.hs
```diff
@@ -0,0 +1,42 @@
...
+import Cardano.Address
+    ( Address )
+
+import Control.Exception
+    ( Exception )
+
+-- $overview
+--
+-- This module provides error/exception types for Addresses.
+
```
KtorZ, line 25: I don't think we can get away with a single type for all `inspectXXX` functions in this module. For example, Jormungandr addresses will never throw an `InvalidDerivationPath` error and, "pure Shelley" addresses won't either. So it may be better to distinguish between Shelley and Byron errors (and Jormungandr on a third type).

core/lib/Cardano/Address/Style/Icarus.hs
```diff
@@ -311,16 +315,20 @@ deriveAddressPublicKey =
...
+inspectIcarusAddress :: MonadThrow m => Address -> m Json.Value
 inspectIcarusAddress addr = do
-    payload <- CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes
-    (root, attrs) <- CBOR.deserialiseCbor decodePayload payload
-    guard $ 1 `notElem` (fst <$> attrs)
-    ntwrk <- CBOR.deserialiseCbor CBOR.decodeProtocolMagicAttr payload
+    payload <- either (throwM . DeserialiseError addr) pure
+            $ CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes
+    (root, attrs) <- either (throwM . DeserialiseError addr) pure
+                  $ CBOR.deserialiseCbor decodePayload payload
```
KtorZ, line 327: Avoid variable-length indentation, there's no need to indent the `$` more than necessary, see https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#avoid-variable-length-indentation

core/lib/Cardano/Address/Style/Icarus.hs
```diff
@@ -311,16 +315,20 @@ deriveAddressPublicKey =
...
+    _  <- maybe (throwM $ InvalidDerivationPath addr) pure
+       $ find ((== 1) . fst) attrs
```
KtorZ, line 329: This actually changes the semantic of the function. Before, this was checking that there was no derivation path present in the address attributes (that's how we distinguish legacy Byron and Icarus addresses). Now, it does the opposite.

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -490,25 +495,24 @@ inspectShelleyAddress addr
...
                         , "stake_reference"   .= Json.String "none"
                         , "script_hash"       .= base16 (BS.take scriptHashSize rest)
                         , "network_tag"       .= network
                         ]
                -- 1000: byron address
                 0b10000000 ->
-                    inspectByronAddress addr <|> inspectIcarusAddress addr
-                _ ->
-                    Nothing
+                  inspectByronAddress addr <|> inspectIcarusAddress addr
```
KtorZ, line 514: I wonder how this plays in plain IO with regard to the `MonadThrow` constraint. I didn't even know there was an `Alternative` instance for IO and at this stage, I am a bit puzzled about which intuition I'd have for it :thinking:

core/lib/Cardano/Address/Errors.hs
hasufell: Sure, since it's MonadThrow, that's easily possible. My initial idea was just for the cli to print better errors. If someone wants to depend on the constructors for control flow, then I guess we need to distinguish.

core/lib/Cardano/Address/Errors.hs
hasufell: > It is generally not a good idea to manually define custom stock instances like

There is no way around it if there is an existential type inside

KtorZ: What about using `StandaloneDeriving` and simply:

```hs
{-# LANGUAGE StandaloneDeriving #-}

deriving instance Show AddrError
```

with an adequate `Show` constraint on `e` ?

hasufell: Interesting, I didn't know about that trick. The problem is also existential types like these violate the haskell standard, so stock instances shouldn't work, afaiu.

core/lib/Cardano/Address/Style/Shelley.hs
hasufell: MonadThrow + IO always ends up es exceptions. Alternative in IO will then catch them and try the other branch.

KtorZ: Sweet.

hasufell: Actually, not true :cry: ... MonadThrow + IO results in Exceptions, but these don't specialize to `IOException`... and `(<|>)` only catches those.

command-line/lib/Command/Address/Inspect.hs
```diff
@@ -24,6 +28,10 @@ import Cardano.Address.Style.Shelley
     ( inspectShelleyAddress )
 import Control.Applicative
     ( (<|>) )
+import Control.Exception
```
KtorZ, line 31: Please no implicit imports of modules, this rapidly becomes a mess otherwise :pray:
 
https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#we-use-explicit-imports-by-default-and-favor-qualified-imports-for-ambiguous-functions

command-line/lib/Command/Address/Inspect.hs
```diff
@@ -24,6 +28,10 @@ import Cardano.Address.Style.Shelley
...
+import Control.Monad.Error.Class
+    ( Error )
+import Data.Bifunctor
```
KtorZ, line 34: Same here :point_up:

core/lib/Cardano/Address/Style/Byron.hs
```diff
@@ -289,16 +293,20 @@ deriveAddressPrivateKey acctK =
...
-    payload <- CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes
-    (root, attrs) <- CBOR.deserialiseCbor decodePayload payload
-    path  <- find ((== 1) . fst) attrs
-    ntwrk <- CBOR.deserialiseCbor CBOR.decodeProtocolMagicAttr payload
+    payload <- either (throwM . BrDeserialiseError) pure
+      $ CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes
+    (root, attrs) <- either (throwM . BrDeserialiseError) pure
+      $ CBOR.deserialiseCbor decodePayload payload
+    path  <- maybe (throwM BrInvalidDerivationPath) pure
+      $ find ((== 1) . fst) attrs
```
KtorZ, line 307: Small note: the issue here isn't that the path is invalid, but that the path is missing / not present in the address attributes. I'd suggest something along the lines of `MissingExpectedDerivationPath`

core/lib/Cardano/Address/Style/Icarus.hs
```diff
@@ -311,16 +313,21 @@ deriveAddressPublicKey =
...
-    (root, attrs) <- CBOR.deserialiseCbor decodePayload payload
-    guard $ 1 `notElem` (fst <$> attrs)
-    ntwrk <- CBOR.deserialiseCbor CBOR.decodeProtocolMagicAttr payload
+    payload <- either (throwM . IcDeserialiseError) pure
+      $ CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes
+    (root, attrs) <- either (throwM . IcDeserialiseError) pure
+      $ CBOR.deserialiseCbor decodePayload payload
+    _  <- if (notElem 1 . fmap fst) attrs
+             then pure ()
+             else throwM IcInvalidDerivationPath
```
KtorZ, line 328: What about:

```hs
when (1 `elem` fst <$> attrs) $ 
    throwM IcInvalidDerivationPath
```

Also, I'd suggest renaming this to `UnexpectedDerivationPath` for this is the source of the problem here. We do not expected Icarus addresses to have a derivation path attributes.

core/lib/Cardano/Address/Errors.hs
```diff
@@ -0,0 +1,61 @@
...
+module Cardano.Address.Errors
+    ( -- $overview
+
+      -- * AddrError
+      ShelleyAddrError(..)
+    , ByronAddrError(..)
+    , IcarusAddrError(..)
+    , JormungandrAddrError(..)
+  ) where
```
KtorZ, line 16: Wouldn't it be better to have errors specific to each modules inside the module they relate to (or close)? These errors share nothing in common apart from the fact that they are errors. If there was a common and abstract "Error" data-structure that had some well-defined properties / interfaces that would indeed deserve a separate module. Here however, it seems to me that errors are moved in a separate module for no apparent reasons :thinking: ?

command-line/lib/Command/Address/Inspect.hs
```diff
@@ -64,15 +73,26 @@ mod liftCmd = command "inspect" $
...
+
 run :: Cmd -> IO ()
 run Inspect = do
     bytes <- hGetBytes stdin
-    case inspect (unsafeMkAddress bytes) of
-        Nothing   -> fail "Unrecognized address on standard input"
-        Just json -> BL8.hPutStrLn stdout (Json.encodePretty json)
+    case first show $ inspect (unsafeMkAddress bytes) of
+      Right json -> BL8.hPutStrLn stdout (Json.encodePretty json)
+      Left  str  -> fail str
```
KtorZ, line 85: We shouldn't use `Show` instances for user-facing outputs. It'd be better that the errors are rendered properly using a human-readable formatting. `Show` is good for debugging, but not so much for throwing back in users' face ^^" 

There are many pretty-printing libraries available, we use [`fmt`](https://hackage.haskell.org/package/fmt-0.6.1.1/docs/Fmt.html) in `cardano-wallet` (not sure if I like it more than any other but it does the job). So suggestions are welcome too here but we definitely need something that captures the "human readable" aspect of errors that is different from debug Show.

command-line/lib/Command/Address/Inspect.hs
hasufell: I wish there was a tool that did that for me.

KtorZ: There has been attempts in Haskell. I saw some recent very promising plugins for VSCode using language servers, so hopefully portable to many other editors :crossed_fingers:

core/lib/Cardano/Address/Errors.hs
hasufell: I have no strong opinion on that. However, data type modules are often used to avoid circular imports. I naturally prefer to keep data out of function modules. Seeing all at once also enables me to easily see what structure is shared.

command-line/lib/Command/Address/Inspect.hs
hasufell: I don't have positive experience with pretty printing libraries, they're often overdesigned and awkward to use. I like to use quasiquoters for this kind of stuff, but they will break your tags files :p 

Not really a strong opinion here either.

KtorZ: So let's just go for `fmt`, the API is simple and the dependencies reasonable.

hasufell: Note on fmt: it uses OverloadedStrings afais, which is problematic for ByteString (uses Char8). So unicode will not show up correctly. But that's not a problem with String/Text.

hasufell: This turns out to be a little more complex with MonadThrow, since it degrades to `SomeException`, which only has `Show`. We cannot recover it, unless we are in IO and catch for those specific exception types, afais. I think it's fine to abuse Show for this, but ymmv. This was initially done for debugging, which Show is for. Ofc it doesn't hold the `read . show == id` property, but the haskell report only demands that this holds for derived instances afais.

core/lib/Cardano/Address/Errors.hs
```diff
@@ -0,0 +1,91 @@
...
+
+instance Exception ShelleyAddrError
+
+prettyShError :: ShelleyAddrError -> String
+prettyShError ShUnknownAddrType = "Unknown address type"
+prettyShError (ShWrongInputSize i) = format "Wrong input size of {}" i
+prettyShError (ShPtrRetrieveError s) = format "Failed to retrieve pointer (underlying errors was: {})" s
+
+instance Show ShelleyAddrError where
+  show = prettyShError
```
KtorZ, line 49: :upside_down_face: `Show` instances should not be written by hand; `fmt` provides a `Buildable` instance for human-readable formatting! The purpose of `Show` is for debug only.

core/lib/Cardano/Address/Errors.hs
```diff
@@ -0,0 +1,93 @@
...
+    = BrMissingExpectedDerivationPath
+    | forall e . (Exception e, Show e) => BrDeserialiseError e
+
+deriving instance Show ByronAddrError
+
+instance Exception ByronAddrError where
+  displayException = prettyBrError
+
+prettyBrError :: ByronAddrError -> String
+prettyBrError BrMissingExpectedDerivationPath = "Missing expected derivation Path"
```
KtorZ, line 63: unexpected capitalized `Path` here :point_up:

core/lib/Cardano/Address/Style/Icarus.hs
```diff
@@ -311,16 +315,19 @@ deriveAddressPublicKey =
...
-    (root, attrs) <- CBOR.deserialiseCbor decodePayload payload
-    guard $ 1 `notElem` (fst <$> attrs)
-    ntwrk <- CBOR.deserialiseCbor CBOR.decodeProtocolMagicAttr payload
+    payload <- either (throwM . IcDeserialiseError) pure
+      $ CBOR.deserialiseCbor CBOR.decodeAddressPayload bytes
+    (root, attrs) <- either (throwM . IcDeserialiseError) pure
+      $ CBOR.deserialiseCbor decodePayload payload
+    when (elem 1 . fmap fst $ attrs) $ throwM IcUnexpectedDerivationPath
+    ntwrk <- either (throwM . IcDeserialiseError) pure
+      $ CBOR.deserialiseCbor CBOR.decodeProtocolMagicAttr payload
```
KtorZ, line 330: side-note, your editor should be configured to pick up the configuration at the root of the repository:  https://github.com/input-output-hk/cardano-addresses/blob/master/.editorconfig

This specifies, among other things, that we use 4-space indentation for Haskell code. Make sure your editor is correctly configured :+1:

KtorZ (approved):

PR #71, opened 2020-09-11
allow decrypting derivation path of random addresses in 'inspect'

KtorZ:
This introduces a new option '--root', which, when provided, attempts
  to decrypt the derivation path payload of Byron addresses in the
  'inspect' command.

  The commit also revise the 'bootstrap' command to be more consistent
  with other commands, in particular:

  - In 'bootstrap', the path is now given as an optional argument
    (instead of an optional option).

  - In 'bootstrap', the root public key is now passed as an optional
    option (instead of an optional argument).

---

```
$ cat root.pub
xpub18amv7cs8kj0mxpk0l3vk2w6g22vyf7y5texr9huevqg9kd3davgv5j52xrfcf90kxx2zdrrl826pzc2kptgwegzzzpfgddwqkrk2gpc80hdq2

$ cat addr.pub
xpub1kvz64d7yggggk5uc8kf8t8jjmh3djlx7ksr2xu25na5ypjzjs59j8ym5gqga8un7yg8e6et6sex8kx0cejwjtz8gh8pj0zg7kc53nuq4x3xaw

$ cardano-address address bootstrap --root $(cat root.pub) --network-tag 1097911063 0H/0H <<< $(cat addr.pub) | tee addr                         
37btjrVyb4KGCX8gfLx8FCS7eVn4TySHRVbswRTpfMoP6puZfN1eYrZ2LcVMd3oGgivdHiHx58gYESgihZpGdeqRa8nmeJPahWSyFerjxwhoygDk3d

$ cardano-address address inspect <<< $(cat addr)
{
    "stake_reference": "none",
    "address_style": "Byron",
    "address_root": "f9036c28c6ffe6d663d544ecf67049b4ceab788df3326e9a83f7cdcb",
    "derivation_path": "581c93b458439e62dbabbaa18dcd1a7db5fa5e63d6782c03f58bd8603bb6",
    "network_tag": 1097911063
}

$ cardano-address address inspect --root $(cat root.pub) <<< $(cat addr)
{
    "stake_reference": "none",
    "address_style": "Byron",
    "address_root": "f9036c28c6ffe6d663d544ecf67049b4ceab788df3326e9a83f7cdcb",
    "derivation_path": {
        "account_index": "0H",
        "address_index": "0H"
    },
    "network_tag": 1097911063
}

$ cardano-address address inspect --root $(cat addr.pub) <<< $(cat addr)
Error: Failed to decrypt derivation path

$ cardano-address address inspect --root $(cat addr.prv) <<< $(cat addr)
option --root: Failed to convert bytes into a valid extended public key.
```

### DIFF
--- a/command-line/lib/Command/Address/Inspect.hs
+++ b/command-line/lib/Command/Address/Inspect.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeFamilies #-}
@@ -18,16 +19,14 @@ import Prelude hiding
 
 import Cardano.Address
     ( Address, unsafeMkAddress )
-import Cardano.Address.Style.Byron
-    ( inspectByronAddress )
-import Cardano.Address.Style.Icarus
-    ( inspectIcarusAddress )
+import Cardano.Address.Derivation
+    ( XPub )
 import Cardano.Address.Style.Jormungandr
     ( inspectJormungandrAddress )
 import Cardano.Address.Style.Shelley
     ( inspectShelleyAddress )
 import Control.Applicative
-    ( (<|>) )
+    ( optional, (<|>) )
 import Control.Exception
     ( SomeException, displayException )
 import Control.Monad.Error.Class
@@ -36,6 +35,8 @@ import Fmt
     ( format )
 import Options.Applicative
     ( CommandFields, Mod, command, footerDoc, helper, info, progDesc )
+import Options.Applicative.Derivation
+    ( xpubOpt )
 import Options.Applicative.Help.Pretty
     ( bold, indent, string, vsep )
 import System.Exit
@@ -49,9 +50,9 @@ import qualified Data.Aeson as Json
 import qualified Data.Aeson.Encode.Pretty as Json
 import qualified Data.ByteString.Lazy.Char8 as BL8
 
-
-data Cmd = Inspect
-    deriving (Show)
+newtype Cmd = Inspect
+    { rootPublicKey :: Maybe XPub
+    } deriving (Show)
 
 mod :: (Cmd -> parent) -> Mod CommandFields parent
 mod liftCmd = command "inspect" $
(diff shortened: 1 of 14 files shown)


command-line/lib/Command/Address/Inspect.hs
```diff
@@ -95,8 +99,6 @@ run Inspect = do
...
     inspect :: (e ~ SomeException) => Address -> Either e Json.Value
     inspect addr = do
-      foldr1 (<|>)
-        [inspectByronAddress       addr
-        ,inspectIcarusAddress      addr
-        ,inspectJormungandrAddress addr
-        ,inspectShelleyAddress     addr]
+        inspectJormungandrAddress addr
+        <|>
+        inspectShelleyAddress rootPublicKey addr
```
KtorZ, line 104: note: Icarus and Byron types are already included in Shelley addresses.

command-line/test/Command/Address/InspectSpec.hs
```diff
@@ -15,66 +15,83 @@ import Test.Utils
 
 spec :: Spec
 spec = describeCmd [ "address", "inspect" ] $ do
-    specInspectAddress ["Byron", "none"]
+    specInspectAddress ["Byron", "none"] []
         "37btjrVyb4KEgoGCHJ7XFaJRLBRiVuvcrQWPpp4HeaxdTxhKwQjXHNKL4\
         \3NhXaQNa862BmxSFXZFKqPqbxRc3kCUeTRMwjJevFeCKokBG7A7num5Wh"
 
-    specInspectAddress ["Icarus", "none"]
+    specInspectAddress ["Byron", "address_index", "account_index"]
```
hasufell, line 22: Afais we have no unit test for a specific derivation path value? Would that be an interesting thing to have?

core/lib/Cardano/Address/Style/Byron.hs
```diff
@@ -297,21 +298,31 @@ deriveAddressPrivateKey acctK =
...
+
     ntwrk <- either (throwM . BrDeserialiseError) pure
         $ CBOR.deserialiseCbor CBOR.decodeProtocolMagicAttr payload
+
     pure $ Json.object
         [ "address_style"   .= Json.String "Byron"
         , "stake_reference" .= Json.String "none"
         , "address_root"    .= T.unpack (T.decodeUtf8 $ encode EBase16 root)
-        , "derivation_path" .= T.unpack (T.decodeUtf8 $ encode EBase16 $ snd path)
+        , "derivation_path" .= path
```
hasufell, line 325: I think I've been wondering this before... but there's no specification for the json output? Now the derivation path could be encrypted or decrypted. I think we should document that somewhere?

hasufell (approved):

core/lib/Cardano/Address/Style/Byron.hs
KtorZ: I started to think the same. It'd be good to have a json-schema documenting the output indeed (and test checking against that schema?)

hasufell: Another idea would be to have a separate field for the decrypted path?

PR #77, opened 2020-09-28
turn back script hashes as 28 bytes.

KtorZ:
Embargo is still on, but this has been fixed on the node's side already

### DIFF
--- a/command-line/lib/Command/Address/Bootstrap.hs
+++ b/command-line/lib/Command/Address/Bootstrap.hs
@@ -66,7 +66,7 @@ mod liftCmd = command "bootstrap" $
             , indent 4 $ bold $ string $ "| "<>progName<>" key from-recovery-phrase Byron > root.prv"
             , indent 2 $ string ""
             , indent 2 $ bold $ string "$ cat root.prv \\"
-            , indent 4 $ bold $ string $ "| "<>progName<>" key child --legacy 14H/42H > addr.prv \\"
+            , indent 4 $ bold $ string $ "| "<>progName<>" key child --legacy 14H/42H | tee addr.prv \\"
             , indent 4 $ bold $ string $ "| "<>progName<>" key public \\"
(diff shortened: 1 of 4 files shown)


piotr-iohk (approved):

command-line/lib/Command/Address/Bootstrap.hs
```diff
@@ -66,7 +66,7 @@ mod liftCmd = command "bootstrap" $
             , indent 4 $ bold $ string $ "| "<>progName<>" key from-recovery-phrase Byron > root.prv"
             , indent 2 $ string ""
             , indent 2 $ bold $ string "$ cat root.prv \\"
-            , indent 4 $ bold $ string $ "| "<>progName<>" key child 14H/42H > addr.prv"
+            , indent 4 $ bold $ string $ "| "<>progName<>" key child --legacy 14H/42H > addr.prv"
```
piotr-iohk, line 69: Can you add trailing \\ :sweat_smile:

KtorZ: Good point. Done.

piotr-iohk: This example still fails to work, unless I'm doing something wrong...
```
cardano-address recovery-phrase generate --size 12 \
    | cardano-address key from-recovery-phrase Byron > root.prv

cat root.prv \
>     | cardano-address key child --legacy 14H/42H > addr.prv \
>     | cardano-address key public \
>     | cardano-address address bootstrap --root $(cat root.prv | cardano-address key public) \
>         --network-tag 764824073 14H/42H

user error (Couldn't convert bytes into extended private key.)
user error (Couldn't convert bytes into extended public key.)
```

KtorZ: >      | cardano-address key child --legacy 14H/42H > addr.prv \

should be 

```
>      | cardano-address key child --legacy 14H/42H | tee addr.prv \
```

If the output is redirected to a file it can't be piped through the next command.

piotr-iohk: fixed -> f779313

Issue #74, opened 2020-09-17
Script address construction


paweljakubas:
# Context

When having a capability to construct verification keys and script we want to have a way to construct multisig-script address for a given network tag

# Decision

We add a way to construct multisig address for a given script and network tag properly formatted.
# Acceptance Criteria

1. Address with script as stake or payment credential can be constructed
2. Better script error reporting
3. Use **script_hash** hrp in bech32 for script hash used in either payment/stake credential

---

# Development 

# QA 

See tests added, also CLI support added with tests. would be worth to play with it. The number of comments with code snippets also were updated.

piotr-iohk:
LGTM.

> See tests added, also CLI support added with tests. would be worth to play with it.

I played, it was worth it. :)

PR #79, opened 2020-10-02
script addresses

paweljakubas:
#74 

- [x] introduce ScriptError
- [x] move empty list responsibility from Parser to validator
- [x] extend validate errors in order to better suggest user of the error
- [x] move key length checking responsibility from Parser to validator
- [x] add error of duplicate verification keys in a given level
- [x] allow using various formats during script construction
- [x] introduce PaymentCredential and StakeCredential
- [x] enable paymentAddress to use PaymentCredential
- [x] enable extendAddress to use PaymentCredential and StakeCredential
- [x] enable stakeAddress to use StakeCredential
- [x] enable pointerAddress to use PaymentCredential
- [x] enable delegationAddress to use PaymentCredential and StakeCredential
- [x] add CLI support/testing
- [x] update documentation, code snippets


paweljakubas:
I have two failures of test at this moment : 
```
cardano-addresses-cli>     fails if a script is malformed FAILED [1]
cardano-addresses-cli>     fails if a script is malformed FAILED [2]
```
which I do not understand. `ScriptParser` spews `ErrorCall` when this kind of stuff is in input of it. I hoped that this https://github.com/input-output-hk/cardano-addresses/pull/79/files#diff-9af2a1e1b20742c8d9e1738978cd4ff4R32 would be returned. but it is not. it seems https://github.com/input-output-hk/cardano-addresses/pull/79/files#diff-92cbf0077f38a73ecc103bea0b00efcbR64 is triggered : 
```
cardano-addresses-cli> , fails if a script is malformed
cardano-addresses-cli>        "user error (The hash of verification key is expected to have 28 bytes.)\n" does not contain "Parsing of the script failed. The script should be composed of nested lists, and  the verification keys should be either base16 or base58 or bech32 encoded."            
```
so it seems ScriptParser ejects some bytes and then they are tried and not validated as they do not satisfy 28-byte length

rvl:
I fixed the test and moved `scriptFromString` into the parser module.

`describeCmd` makes it difficult to see what the actual test failures are.

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -683,48 +714,63 @@ data ErrExtendAddress
...
+pointerAddress discrimination credential pointer =
+    case credential of
+        PaymentFromKey keyPub ->
+            -- pointer address: keyhash28, 3 variable length uint : 01000000 -> 64
+            Internal.pointerAddress discrimination keyPub pointer
+        PaymentFromScript scriptHash ->
+           -- pointer address: scripthash32, 3 variable length uint : 01010000 -> 80
+            unsafeFromRight $ extendAddress
+            (paymentAddress discrimination (PaymentFromScript scriptHash))
+            (Right pointer)
```
KtorZ, line 754: Why aren't credentials passed to `pointerAddress` directly ? (and same question for `paymentAddress` actually).. 
Wouldn't be easier to modify the signature of the classes above or, to not have the type-classes at actually. 

I mean, in the end, the type-classes would be nice if we wanted to treat credentials and addresses using one single interface but in the end, application code often treats Byron addresses differently from Shelley addresses _anyway_. So there's not much value in capturing this in classes at all.

core/lib/Cardano/Script.hs
```diff
@@ -51,30 +55,28 @@ import qualified Data.List as L
...
 instance NFData Script
 
--- | A 'KeyHash' type represents verification key hash
--- that participate in building MultisigScript. The hash is expected to have size of
+-- | A 'ScriptHash' type represents script hash. The hash is expected to have size of
 -- 28-byte.
 --
 -- @since 3.0.0
-newtype KeyHash = KeyHash ByteString
+newtype ScriptHash = ScriptHash ByteString
```
KtorZ, line 69: Alternatively, we could reuse the `Depth` data-kind and have a `Hash (purpose :: Depth)` data-type. That'd be quite consistent with the approach for keys. So you'd have:

```hs
Shelley 'AddressK XPub -> Hash 'AddressK 
Shelley 'StakingK XPub -> Hash 'StakingK
```

core/lib/Cardano/Script.hs
```diff
@@ -87,15 +89,59 @@ keyHashFromBytes bytes
...
+        Left EmptyList
+    scanContent content
+validateScript (RequireMOf m content) = do
+    when (m == 0) $
+        Left MZero
+    when (length content < fromInteger (toInteger m) ) $
+        Left ListTooSmall
+    scanContent content
+
+scanContent :: [Script] -> Either InvalidScriptError ()
```
KtorZ, line 112: Isn't that simply `traverse validateScript` ?

core/lib/Cardano/Script.hs
```diff
@@ -87,15 +89,59 @@ keyHashFromBytes bytes
...
+-- | Errors when handling 'Script'
+--
+-- @since 3.0.0
+data ScriptError = MalformedScript | InvalidScript InvalidScriptError
+    deriving Eq
+
+data InvalidScriptError = EmptyList | ListTooSmall | MZero | DuplicateSignatures | WrongKeyHash
+    deriving (Eq, Show)
+
+instance Show ScriptError where
```
KtorZ, line 135: Do :pray: 
Not :pray: 
Manually :pray: 
Define :pray: 
Show :pray: 
Instances :pray: 

^^" ... `Show` is really for _debugging_ and showing things in a Haskelly way. To get a human-readable representation of some data-type, it is better to define an explicit function like `prettyThis` or `toStringThat` or, use a type-class that convey this idea (like `Buildable` or `Pretty`)

core/lib/Cardano/Script.hs
```diff
@@ -138,8 +184,8 @@ encodeFoldable encode xs = wrapArray len contents
 -- `Api.serialiseToRawBytes $ Api.scriptHash script` realizes
 -- This is basically doing the symbolically the following (using symbols from toCBOR):
 -- digest $ (nativeMultisigTag <> toCBOR multisigScript )
-toScriptHash :: Script -> ByteString
-toScriptHash script = digest $ nativeMultiSigTag <> toCBOR'' script
+toScriptHash :: Script -> ScriptHash
+toScriptHash script = ScriptHash $ digest $ nativeMultiSigTag <> toCBOR'' script
```
KtorZ, line 188: I am not sure to understand this `nativeMultisigTag` and `toCBOR''` 

Is this to abide by the ad-hoc format used by cardano-cli ? If that's the case, then I think it is wrong. The script hash should be only the blake2b_224 hash of it. We _may_ want to have some compatibility layer with cardano-cli at some point but it shouldn't be the default.

core/lib/Cardano/Script.hs
rvl: There is also `displayException` if you're not against adding an `Exception` instance.

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -625,25 +593,69 @@ inspectShelleyAddress mRootPub addr
...
+    | PaymentFromScript ScriptHash
+    deriving (Show, Eq)
+
+-- | In Shelley stake credential can originate from either key or script.
+--
+-- @since 3.0.0
+data StakeCredential =
+      StakeFromKey (Shelley 'StakingK XPub)
+    | StakeFromScript ScriptHash
+    deriving (Show, Eq)
```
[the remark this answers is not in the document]
paweljakubas: good idea, replaced with signle constructor Credential-> https://github.com/input-output-hk/cardano-addresses/pull/79/commits/d0eb6aa0b7eda3997e8305f04d39e203b18b59f2

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -652,19 +664,38 @@ extendAddress addr stakeReference = do
...
+                putByteString scriptBytes
+
+        -- pointer address: keyhash28, 3 variable length uint    : 01000000 -> 64
+        -- pointer address: scripthash32, 3 variable length uint : 01010000 -> 80
         Right pointer -> do
             pure $ unsafeMkAddress $ BL.toStrict $ runPut $ do
-                putWord8 $ fstByte .&. 0b01001111
+                -- 0b01100000 .&. 0b01011111 = 64
+                -- 0b01110000 .&. 0b01011111 = 80
+                putWord8 $ fstByte .&. 0b01011111
```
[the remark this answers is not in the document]
paweljakubas: please. I used https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley/chain-and-ledger/shelley-spec-ledger-test/cddl-files/shelley.cddl as reference plus checked every case by hand in ghci

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -625,25 +593,69 @@ inspectShelleyAddress mRootPub addr
...
+-- | In Shelley stake credential can originate from either key or script.
+--
+-- @since 3.0.0
+data StakeCredential =
+      StakeFromKey (Shelley 'StakingK XPub)
+    | StakeFromScript ScriptHash
+    deriving (Show, Eq)
+
+constructPayload
+    :: Word8
```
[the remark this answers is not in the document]
paweljakubas: addressed in https://github.com/input-output-hk/cardano-addresses/pull/79/commits/c2df3de55ba2bcdbba949cfd6468a4372a9b1751

core/lib/Cardano/Address/Style/Shelley.hs
paweljakubas: agree Pointer/Delegation/Payment impl are redundant here as they capture only subset of arguments -> removed them, needed to relocate some testing from Cardano.AddressesSpec to ShelleySpec as a consequence. Done in https://github.com/input-output-hk/cardano-addresses/pull/79/commits/c2df3de55ba2bcdbba949cfd6468a4372a9b1751

core/lib/Cardano/Script.hs
paweljakubas: I would argue that as ScriptHash is computed based on script which has rulings plus verification key hashes, then maybe it is good as it is. No need to add differentiation. The same ScriptHash can go to payment and stake credential. We captured differentiation already in Credential (purpose::Depth)

core/lib/Cardano/Script.hs
paweljakubas: so I wanted two-step golden test here, namely for (Api stands for cardano-api):
1. Api.serialiseToCBOR script
2. Api.serialiseToRawBytes $ Api.scriptHash script

But there is problem here how cbor is handled. Symbolically,
    Api.serialiseToCBOR script -> toCBOR [0,multisigScript]
    Api.scriptHash script -> digest $ (nativeMultisigTag <> toCBOR multisigScript )
and this is reflected in the toCBOR' and toCBOR''
This is how cardano-node handles it, spent some time debugging that, and could only adjust to it

core/lib/Cardano/Script.hs
paweljakubas: yes ... I had first toText, but then realised we do not have it...and I am used to it. corrected https://github.com/input-output-hk/cardano-addresses/pull/79/commits/f72feec2b54adcf95983289edfae21688f14d051 btw. sometimes I think it would be good to have adrestia-base or adrestia-core repo in which we have all those intensively reusable things like encode/decode, toText, fromText, ... to name a few. Each repo would use it as dependency and there would be a lot of code cutting as a result

command-line/lib/Options/Applicative/Script.hs
```diff
@@ -9,30 +9,77 @@ module Options.Applicative.Script
...
+scriptHashArg helpDoc =
+    argument (eitherReader scriptHashReader) $ mempty
+        <> metavar "SCRIPT_HASH"
+        <> help helpDoc
+
+data StakeCredential = FromKey XPub | FromScriptHash ScriptHash
+    deriving (Show, Eq)
+
+stakeCredentialArg  :: String -> Parser StakeCredential
+stakeCredentialArg str = (FromScriptHash <$> scriptHashArg str) <|> (FromKey <$> xpubArg str)
```
paweljakubas, line 85: need to do something that when xpub is in str then `Failed to convert bytes into a valid script hash` is not thrown

paweljakubas: fixed in https://github.com/input-output-hk/cardano-addresses/pull/79/commits/1aee6f8ba433eef62452070b75a2fceee978752c

core/lib/Cardano/Script.hs
paweljakubas: probably -> https://twitter.com/impurepics/status/1312158622302924801

command-line/lib/Command/Address/Reward.hs
```diff
@@ -68,15 +70,21 @@ mod liftCmd = command "stake" $
...
         Left e -> die (fmt $ build e)
         Right discriminant -> do
-            let addr = Shelley.stakeAddress discriminant (Shelley.liftXPub xpub)
+            addr <- case credential of
+                    CredentialFromKey -> do
+                        xpub <- hGetXPub stdin
+                        pure $ Shelley.stakeAddress discriminant (Shelley.FromKey $ Shelley.liftXPub xpub)
+                    CredentialFromScript -> do
+                        scriptHash <- hGetScriptHash stdin
+                        pure $ Shelley.stakeAddress discriminant (Shelley.FromScript scriptHash)
```
KtorZ, line 87: For consistency, wouldn't it be better that we accept a _script_ has an input here, and simply store it as a hash ?

command-line/lib/Options/Applicative/Script.hs
```diff
@@ -9,30 +9,79 @@ module Options.Applicative.Script
...
+
+scriptHashArg :: String -> Parser ScriptHash
+scriptHashArg helpDoc =
+    option (eitherReader scriptHashReader) $ mempty
+        <> long "from-script"
+        <> metavar "SCRIPT_HASH"
+        <> help helpDoc
+
+data StakeCredential = FromKey XPub | FromScriptHash ScriptHash
+    deriving (Show, Eq)
```
KtorZ, line 83: What is specific to `Stake` here ? How's different from any other key/script credential ?

command-line/lib/Options/Applicative/Script.hs
```diff
@@ -9,30 +9,79 @@ module Options.Applicative.Script
...
+
+scriptHashReader :: String -> Either String ScriptHash
+scriptHashReader str = do
+    bytes <- case detectEncoding str of
+        Just EBase16   -> fromBase16 (toBytes str)
+        Just EBech32{} -> fromBech32 markCharsRedAtIndices (toBytes str)
+        Just EBase58   -> fromBase58 (toBytes str)
+        Nothing        -> Left
+            "Couldn't detect input encoding? The key must be encoded as \
+            \base16, bech32 or base58."
```
KtorZ, line 67: It seems that we are doing this in a few places, maybe there's some room for factorization to avoid duplication?

core/lib/Cardano/Script.hs
KtorZ: But cardano-api has its own ad-hoc format for serializing things which is not actually matching the on-chain binary representation (but is a little wrapper around it). I don't think we should bother here and should treat ScriptHash as plain bytes like they are. The fact that cardano-api encodes them as CBOR with a specific tag is superfluous _at this level_. 

Later on, I am all for reconciating both interfaces in one common format, or have ways to convert from the "plain" bytes to cardano-api-specific bytes but arguably, that is something which should live in `cardano-cli`, not in here.

command-line/lib/Options/Applicative/Script.hs
paweljakubas: indeed, reused Credential 'StakingK -> https://github.com/input-output-hk/cardano-addresses/pull/79/commits/050f2ed0e944c2e1fc7ab0bc5dfa7747476ddddc

command-line/lib/Options/Applicative/Script.hs
paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/79/commits/49268acd7873061f1ff1e84b732e7220bf704fef

core/lib/Cardano/Script.hs
paweljakubas: as I underlined. I included this as I tried to do it in steps, first replicate toCBOR, then scritpHash calculations. It tutn out later that cardano-node has very surprising impl of scriptHash. Agree that it should be in one place - `adrestia-core` :-) and then they should use it in other repos:-)

paweljakubas: also please notice that I used this PR (here because of dependiencies) https://github.com/input-output-hk/cardano-transactions/pull/23 to get cbor and script hash values

rvl:
It looks like the stylish-haskell step of the "Coding standards" build fails because it is not using the `.stylish-haskell.yaml` config file in this repo.

KtorZ (approved):
reviewed and pushed some house-keeping / adjustment

PR #84, opened 2020-10-14
to/from json instances for script

paweljakubas:
https://jira.iohk.io/browse/ADP-474

```
"e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a" 
{ "all" : [ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a"
          , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735b"
          ]
}
{ "all" : [ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a"
          , {"any": [ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735b"
                    ,  "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735c"
                    ] 
            }
          ]
}
{ "all" : [  "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a"
          , {"some": { "from" :[ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735b"
                      , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735c"
                      , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735d"
                      ] 
                        , "at_least" : 2
                     }
            }
          ]
}
```

- [x] ToJSON for Script
- [x] FromJSON for Script
- [x] testing/roundtrips
- [x] all hrp in line with https://github.com/cardano-foundation/CIPs/pull/31

### DIFF
--- a/core/lib/Cardano/Address/Script.hs
+++ b/core/lib/Cardano/Address/Script.hs
@@ -36,7 +36,7 @@ module Cardano.Address.Script
 import Prelude
 
 import Codec.Binary.Encoding
-    ( AbstractEncoding (..), detectEncoding, encode, fromBase16, fromBase58 )
+    ( AbstractEncoding (..), detectEncoding, encode, fromBase16 )
 import Control.Applicative
     ( (<|>) )
 import Control.DeepSeq
@@ -241,10 +241,11 @@ showErr (ErrorKeyHashFromTextInvalidString EBase16) =
 showErr (ErrorKeyHashFromTextInvalidString EBech32{}) =
     "Invalid Bech32-encoded string."
 showErr (ErrorKeyHashFromTextInvalidString EBase58) =
-    "Invalid Base58-encoded string."
+    "Verification key hash must be must be encoded as \
+    \base16 or bech32."
 showErr ErrorKeyHashFromTextWrongEncoding =
     "Verification key hash must be must be encoded as \
-    \base16, bech32 or base58."
+    \base16 or bech32."
 showErr ErrorKeyHashFromTextWrongPayload =
     "Verification key hash must contain exactly 28 bytes."
 showErr ErrorKeyHashFromTextWrongHrp =
@@ -263,9 +264,7 @@ keyHashFromText txt =  case detectEncoding str of
             Left _ -> Left $ ErrorKeyHashFromTextInvalidString EBase16
             Right bytes -> checkPayload bytes
         Just EBech32{} -> fromBech32
-        Just EBase58 -> case fromBase58 (toBytes str) of
-            Left _ -> Left $ ErrorKeyHashFromTextInvalidString EBase58
-            Right bytes -> checkPayload bytes
+        Just EBase58 -> Left ErrorKeyHashFromTextWrongEncoding
         Nothing -> Left ErrorKeyHashFromTextWrongEncoding
  where
     str = T.unpack txt
--- a/core/lib/Cardano/Address/Script.hs
+++ b/core/lib/Cardano/Address/Script.hs
@@ -94,7 +94,7 @@ data Script
     = RequireSignatureOf !KeyHash
     | RequireAllOf ![Script]
     | RequireAnyOf ![Script]
-    | RequireMOf Word8 ![Script]
+    | RequireSomeOf Word8 ![Script]
     deriving stock (Generic, Show, Eq)
 instance NFData Script
 
@@ -116,7 +116,7 @@ validateScript = \case
         when (hasDuplicate script) $ Left DuplicateSignatures
         traverse_ validateScript script
 
-    RequireMOf m script -> do
+    RequireSomeOf m script -> do
         when (m == 0) $ Left MZero
         when (length script < fromIntegral m) $ Left ListTooSmall
         when (hasDuplicate script) $ Left DuplicateSignatures
@@ -171,7 +171,7 @@ serialize script =
             encodeMultiscriptCtr 1 2 <> encodeFoldable toCBOR contents
         RequireAnyOf contents ->
             encodeMultiscriptCtr 2 2 <> encodeFoldable toCBOR contents
-        RequireMOf m contents -> mconcat
+        RequireSomeOf m contents -> mconcat
             [ encodeMultiscriptCtr 3 3
             , CBOR.encodeInt (fromInteger $ toInteger m)
             , encodeFoldable toCBOR contents
@@ -319,12 +319,12 @@ bech32 (KeyHash keyHash) = T.decodeUtf8 $ encode (EBech32 hrp) keyHash
 --          ]
 --}
 --{ "all" : [ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a"
---          , {"at_least": { "from" :[ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735b"
---                                   , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735c"
---                                   , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735d"
---                                   ]
---                         , "m" : 2
---                         }
+--          , {"some": { "from" :[ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735b"
+--                               , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735c"
+--                               , "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735d"
+--                               ]
+--                     , "at_least" : 2
+--                     }
 --            }
 --          ]
 --}
@@ -335,9 +335,9 @@ instance ToJSON Script where
         object ["all" .= fmap toJSON content]
     toJSON (RequireAnyOf content) =
         object ["any" .= fmap toJSON content]
-    toJSON (RequireMOf m content) =
-        let inside = Object ("from" .= fmap toJSON content <> "m" .= toJSON m)
-        in object ["at_least" .= inside]
+    toJSON (RequireSomeOf m content) =
+        let inside = Object ("from" .= fmap toJSON content <> "at_least" .= toJSON m)
+        in object ["some" .= inside]
 
 instance FromJSON Script where
     parseJSON v = parseKey v <|> parseAnyOf v  <|> parseAllOf v <|> parseAtLeast v
@@ -348,8 +348,8 @@ instance FromJSON Script where
               withObject "Script AnyOf" $ \o -> RequireAnyOf <$> (o .: "any" :: Json.Parser [Script])
           parseAllOf =
               withObject "Script AllOf" $ \o -> RequireAllOf <$> (o .: "all" :: Json.Parser [Script])
-          parseAtLeast = withObject "Script MOf" $ \o -> do
-              obj <- o .: "at_least"
+          parseAtLeast = withObject "Script SomeOf" $ \o -> do
+              obj <- o .: "some"
               content <- obj .: "from"
-              m <- obj .: "m"
-              RequireMOf m <$> parseJSON content
+              m <- obj .: "at_least"
+              RequireSomeOf m <$> parseJSON content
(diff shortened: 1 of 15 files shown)


core/lib/Cardano/Script.hs
```diff
@@ -212,3 +220,68 @@ blake2b224 =
 -- Size, in bytes, of a hash of public key (without the corresponding chain code)
 hashSize :: Int
 hashSize = hashDigestSize Blake2b_224
+
+-- Examples of Script jsons:
+--{ "key" : "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a" }
```
rvl, line 225: Rather than a json object, could this just be a string?
The label "key" is not quite right because it's a hex-encoded hash of a key.
And were we going to use bech32 for key hashes?

core/test/Cardano/ScriptSpec.hs
```diff
@@ -251,6 +261,34 @@ spec = do
...
+            let roundtrip =
+                    Json.decode . Json.encode . toJSON :: Script -> Maybe Script
+            property $ \(t :: Script) -> roundtrip t === pure t
+
   where
     toHexText = T.decodeUtf8 . encode EBase16
     toHexText' (ScriptHash bytes) = toHexText bytes
+
+instance Arbitrary Script where
+    arbitrary = do
```
rvl, line 276: I can't remember ... with generators of recursive types, does the generator need to be made smaller for every step down the tree?

rvl (commented):
Looks ok.

Is this schema specified anywhere? Do we need to have a discussion about the schema?

core/lib/Cardano/Script.hs
```diff
@@ -31,6 +31,8 @@ module Cardano.Script
 
 import Prelude
 
+import Codec.Binary.Encoding
+    ( AbstractEncoding (..), encode, fromBase16 )
```
rvl, line 35: I forgot to add that I don't really like module names like this.
It confuses me because it looks like the module comes from another package.
Given that it deals with bech32/base56, it is very much a module which should be called `Cardano.Addresses.Encoding`.

Also, this module `Cardano.Script`. There are multiple kinds of scripts in cardano. How about `Cardano.Addresses.Script`?

core/lib/Cardano/Script.hs
```diff
@@ -212,3 +220,68 @@ blake2b224 =
...
+--                                   ]
+--                         , "m" : 2
+--                         }
+--            }
+--          ]
+--}
+instance ToJSON Script where
+    toJSON (RequireSignatureOf (KeyHash key)) =
+        let keyHexed = T.decodeUtf8 $ encode EBase16 key
+        in Object ("key" .= keyHexed)
```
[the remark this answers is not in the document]
paweljakubas: done in https://github.com/input-output-hk/cardano-addresses/pull/84/commits/3c2bb9a1e8743c88bddb7ab271bce8fe393eea09

core/lib/Cardano/Script.hs
paweljakubas: good idea, I was also thinking about it (ie. to express verification key hash as string). also in accordance to "line parser" added all, bech32, base16 and base58, support. see https://github.com/input-output-hk/cardano-addresses/pull/84/commits/3c2bb9a1e8743c88bddb7ab271bce8fe393eea09

core/lib/Cardano/Script.hs
```diff
@@ -212,3 +220,68 @@ blake2b224 =
...
+        reqAny <-
+            (withObject "script" $
+             \o -> o .:? "any" :: Json.Parser (Maybe [Script])) obj
+        reqAll <-
+            (withObject "script" $
+             \o -> o .:? "all" :: Json.Parser (Maybe [Script])) obj
+        mOfN <-
+            (withObject "script" $
+             \o -> o .:? "at_least" :: Json.Parser (Maybe Value)) obj
+        case (reqKey, reqAny, reqAll, mOfN) of
```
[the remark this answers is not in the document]
paweljakubas: done https://github.com/input-output-hk/cardano-addresses/pull/84/commits/3c2bb9a1e8743c88bddb7ab271bce8fe393eea09

core/test/Cardano/ScriptSpec.hs
paweljakubas: :thinking:

core/lib/Cardano/Script.hs
paweljakubas: sound idea -> https://github.com/input-output-hk/cardano-addresses/pull/84/commits/b899a28d6c6cd517bdf854bc7a75b8165cb7a1e2

core/lib/Cardano/Address/Script.hs
```diff
@@ -200,6 +228,62 @@ keyHashFromBytes bytes
...
+-- binary payload is expected to be composed of 28 bytes.
+--
+-- @since 3.0.0
+keyHashFromText :: Text -> Either ErrorKeyHashFromText KeyHash
+keyHashFromText txt =  case detectEncoding str of
+        Just EBase16 -> case fromBase16 (toBytes str) of
+            Left _ -> Left $ ErrorKeyHashFromTextInvalidString EBase16
+            Right bytes -> checkPayload bytes
+        Just EBech32{} -> fromBech32
+        Just EBase58 -> case fromBase58 (toBytes str) of
```
rvl, line 267: I don't think we should support base58 for key hashes because it will just add confusion. I would even be mildly in favour of only using bech32 and not even supporting hex. The thing I like about bech32 is the HRP - which base16 and base58 don't have. On the other hand, supporting hex is good for interoperability.

core/lib/Cardano/Address/Script.hs
```diff
@@ -212,3 +296,60 @@ blake2b224 =
...
+        let inside = Object ("from" .= fmap toJSON content <> "m" .= toJSON m)
+        in object ["at_least" .= inside]
+
+instance FromJSON Script where
+    parseJSON v = parseKey v <|> parseAnyOf v  <|> parseAllOf v <|> parseAtLeast v
+      where
+          parseKey = withText "Script KeyHash" $
+              either (fail . showErr) (pure . RequireSignatureOf) . keyHashFromText
+          parseAnyOf =
+              withObject "Script AnyOf" $ \o -> RequireAnyOf <$> (o .: "any" :: Json.Parser [Script])
```
rvl, line 348: Are the type annotations needed to make it build?

core/test/Cardano/Address/ScriptSpec.hs
```diff
@@ -251,6 +261,34 @@ spec = do
...
+            n <- choose ((fromInteger $ toInteger m),10)
+            pure $ RequireMOf m <$> vector n
+        let reqSig =
+                (RequireSignatureOf . KeyHash . BS.pack) <$> replicateM 28 arbitrary
+        oneof
+            (replicate 15 reqSig ++
+            [ reqAllGen
+            , reqAnyGen
+            , reqMofNGen
+            ])
```
rvl, line 294: I looked it up... We do need to be careful to avoid generators which fail to terminate, or produce very large results. [Generating Recursive Data Types](http://www.cse.chalmers.se/~rjmh/QuickCheck/manual_body.html#16) :thinking:.

Also, can `shrink = genericShrink` work here?

```suggestion

instance Arbitrary Script where
    arbitrary = sized script'
      where
        script' 0 = RequireSignatureOf . KeyHash . BS.pack <$> replicateM 28 arbitrary
        script' n | n>0 = oneof
            [ RequireAllOf <$> listOf arbitrary
            , RequireAnyOf <$> listOf arbitrary
            , RequireMOf <$> choose (2,5) <*> listOf arbitrary
                -- note: perhaps the last one is too invalid
            ]
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -212,3 +296,60 @@ blake2b224 =
...
+
+instance ToJSON Script where
+    toJSON (RequireSignatureOf keyHash) = String $ bech32 keyHash
+    toJSON (RequireAllOf content) =
+        object ["all" .= fmap toJSON content]
+    toJSON (RequireAnyOf content) =
+        object ["any" .= fmap toJSON content]
+    toJSON (RequireMOf m content) =
+        let inside = Object ("from" .= fmap toJSON content <> "m" .= toJSON m)
+        in object ["at_least" .= inside]
```
rvl, line 340: Could we rename `RequireMOf` to `RequireSomeOf` or `RequireAtLeast`? That is more descriptive to me.

Then with the JSON, perhaps `"at_least"` → `"some"` and `"m"` → `"at_least"`.

Also it's probably possible to shorten this:

```suggestion
    toJSON (RequireSomeOf count scripts) =
      object ["some" .= (object ["at_least" .= count, "from" .= scripts])]
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -200,6 +228,62 @@ keyHashFromBytes bytes
...
+    str = T.unpack txt
+    toBytes = T.encodeUtf8 . T.pack
+    checkPayload bytes =
+        maybeToRight ErrorKeyHashFromTextWrongPayload (keyHashFromBytes bytes)
+    fromBech32 = do
+        (hrp, dp) <- either
+            (const $ Left $ ErrorKeyHashFromTextInvalidString $ EBech32 ())
+            Right (Bech32.decodeLenient txt)
+        case Bech32.humanReadablePartToText hrp of
+            "script_vkh" -> do
```
rvl, line 281: Is this specified anywhere?

rvl (approved):
Looks better. I have suggestions for the json schema. It would be good to get the json and bech32 HRPs are correct now so we don't need to fix it later.

core/lib/Cardano/Address/Script.hs
paweljakubas: yes, it it specified here -> https://github.com/cardano-foundation/CIPs/pull/31

core/lib/Cardano/Address/Script.hs
paweljakubas: good idea -> done in https://github.com/input-output-hk/cardano-addresses/pull/84/commits/4b26932a4ea05bacca33d0efab79e2d813d686f0

core/test/Cardano/Address/ScriptSpec.hs
paweljakubas:
```
instance Arbitrary Script where
    arbitrary = sized script'
      where
        script' 0 = RequireSignatureOf . KeyHash . BS.pack <$> replicateM 28 arbitrary
        script' n =
            let n' = n `div` 2
                scripts = listOf (script' n')
                conds = [ RequireAllOf <$> scripts
                        , RequireAnyOf <$> scripts ]
            in if n' >= 5 then
                   oneof (conds ++ [RequireSomeOf <$> choose (2,5) <*> scripts])
               else
                   oneof conds

```
seems to stuck when running... on the other hand, for me, it is much less clear to read

core/lib/Cardano/Address/Script.hs
paweljakubas: indeed - not needed - removed here https://github.com/input-output-hk/cardano-addresses/pull/84/commits/6d001e1c93efe19182d250027aaf32352ba40673

core/lib/Cardano/Address/Script.hs
paweljakubas: now support only to bech32 and base16 (I think some people are very attached to this format - like me;-))  -> https://github.com/input-output-hk/cardano-addresses/pull/84/commits/6a1ebd16d433c16b5d8919743ac9a3873248e612

core/lib/Cardano/Address/Script.hs
rvl: Thanks

core/test/Cardano/Address/ScriptSpec.hs
rvl: OK I fixed it. I think it's important to learn how to generate tests properly and use best practices for quickcheck.

rvl (approved):

PR #83, opened 2020-10-14
Two minor fixes to the cabal.project file

erikd:

PR #85, opened 2020-10-15
Use stack2cabal to generate cabal.project{,freeze}

hasufell:
To generate, run `stack2cabal -p now`.

Related: https://github.com/input-output-hk/cardano-addresses/pull/83

This should make syncing easier.


cabal.project
```diff
@@ -1,9 +1,28 @@
...
+    core/
+  , command-line/
+
+source-repository-package
+    type: git
+    location: https://github.com/input-output-hk/cardano-crypto
+    tag: 2547ad1e80aeabca2899951601079408becbc92c
+
+allow-older: *
+allow-newer: *
```
erikd, line 17: Probably not a good idea. The constraints and specific `allow-newer` were mush more exact and fine grained.

hasufell: All dependencies are pinned in `cabal.project.freeze`. These two lines are necessary to match stacks behavior, which prefers resolver versions over foo.cabal versions. Now the build plan is guaranteed to match the stack build plan though.

erikd: Most of the other projects in the Cardano stack do not use a `cabal.project.freeze` file. The `index-state` probably more than sufficient to solve any dependency version issues.

hasufell: That's also true, yes. It's a different philosophy. stack2cabal is nice to sync stack and cabal. If you want more flexibility with `cabal.project`, then it doesn't make sense.

erikd: I would therefore suggest that the freeze file is not commited to git.

hasufell: Since I'm mantaining stack2cabal currently, I wonder if it would help if i add an option to skip pinning the versions, but still record the flags in cabal.profect.freeze?

erikd: In the presence of `index-state` I see zero value in the `cabal.project.freeze` file.

hasufell: That wouldn't work, because of package flags that need to be pinned.

erikd: Which package flags need to be pinned?

hasufell: > Which package flags need to be pinned?

Those that are pinned in `stack.yaml`.

erikd: The `stack.yaml` file over specifies the dependencies.

hasufell: I don't understand what you mean.

https://github.com/input-output-hk/cardano-addresses/blob/1939121afcd125b6c6300a1b2a8b014007a0356c/stack.yaml#L21-L25

This is supposed to go into the cabal.project/cabal.project.freeze, no?

KtorZ (approved):

PR #86, opened 2020-10-22
Add haskell-language-server config

rvl:
This file is hand-generated because that seems to be the way it's done.

To use it, set up your editor for LSP, and run:

    ln -s hie-cabal.yaml hie.yaml


paweljakubas (approved):
lgtm!

PR #87, opened 2020-10-30
Add pub option for public keys

paweljakubas:
For credentials from public key we need to have public keys without chain code. 
https://github.com/input-output-hk/cardano-wallet/pull/2253

This PR adds a way to get public key (without chain code) if `--pub` is sneaked. Otherwise extended public key is returned.


README.md
```diff
@@ -40,31 +40,42 @@ east student silly already breeze enact seat trade few way online skin grass hum
...
+</details>
+
+<details>
+  <summary>How to generate a public stake key (public key without chain code)</summary>
+
+```
+$ cat recovery-phrase.prv | cardano-address key from-recovery-phrase Shelley \
+| cardano-address key child 1852H/1815H/0H/2/0 \
+| cardano-address key public --pub
+xpub16apaenn9ut6s40lcw3l8v68xawlrlq20z2966uzcx8jmv2q9uy7qh83kg9
```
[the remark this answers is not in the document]
paweljakubas: or more precisely I would like `addr_vk` if we are dealing with payment/delegating address and `stake_vk` if we are dealing with stake address (reward account). If not possible we will need a way to have `--bech32WithHrp prefix`

command-line/lib/Options/Applicative/Public.hs
```diff
@@ -0,0 +1,31 @@
...
+    , publicOpt
+    ) where
+
+import Prelude
+
+import Options.Applicative
+    ( Parser, flag, long )
+
+
+data PublicType = WithChainCode | WithoutChainCode
```
rvl, line 20: Does a single type and parse function need its own module?

command-line/lib/Command/Key/Public.hs
```diff
@@ -40,13 +43,19 @@ mod liftCmd = command "public" $
         <> progDesc "Get the public counterpart of a private key"
         <> footerDoc (Just $ string $ mconcat
             [ "The private key is read from stdin."
+            , "Extended public key is default. In order to get public key"
+            , ", ei., without chain code use '--pub'."
```
rvl, line 47: Do you know which use case is more common?
I am thinking that public key only might be a better default. Not sure though.
Another alternative could be to have completely separate subcommands -- e.g. `public` and `xpub`, or something like that. Then you could tell users to run `xpub` if they need to derive child keys.

README.md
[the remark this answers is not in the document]
rvl: Perhaps use pub/xpub only in cardano-address, and show how to use the bech32 util to change HRP.

README.md
```diff
@@ -40,31 +40,42 @@ east student silly already breeze enact seat trade few way online skin grass hum
   <summary>How to generate a root private key</summary>
 
 ```
-$ cardano-address recovery-phrase generate --size 15 > recovery-phrase.prv
-$ cat recovery-phrase.prv | cardano-address key from-recovery-phrase Shelley
-xprv1fzu4e8cecxshgzzxzh7557sd8tffqreeq2je7fgsm7f02mq849vdupw7qwgxc3qawyqev0l8ew0f4fkp8hvr8mskz4hz6e6ejzjlevcskcl6lqpr07u7552fsfgteztuclse7luh4cp493zdhkrjdss0250cdw8n
+$ cat recovery-phrase.txt | cardano-address key from-recovery-phrase Shelley
```
KtorZ, line 43: Why this change ? I think it was clearer before with both instructions visible in the same excerpt.

README.md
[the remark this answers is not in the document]
KtorZ: What do you think of flipping this around and, have this being the default behavior, but require an explicit `--with-chain-code` flag to return the public key with chain code.
 
We should avoid returning the chain code unless the user knows what he/she is doing with it. Changing the default, we would also need to change a bit the error message when failing to parse bytes into a public key + chain code (so that it hints users in the right direction, that a 32-byte chain code is missing).

command-line/lib/Command/Key/Public.hs
```diff
@@ -40,13 +43,19 @@ mod liftCmd = command "public" $
...
             ])
   where
     parser = Public
         <$> encodingOpt [humanReadablePart|xpub|]
+        <*> publicOpt
```
rvl, line 52: Sometthing like this will let you change the HRP...

```suggestion
  parser = do
    pub <- publicOpt
    Public <$> encodingOpt (if pub == WithChainCode then [humanReadablePart|xpub|] else [humanReadablePart|pub|])
        <*> pure pub
```

(clean it up though)

rvl (commented):
Looks good so far - do you think pub or xpub should be the "default"? Or should there be separate subcommands for pub and xpub?

README.md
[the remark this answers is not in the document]
rvl: I like that idea of requiring the user to specify what kind of key they need. Such as:
-  `(--with-chain-code | --without-chain-code )`.
- `cardano-address key xpub` /  `cardano-address key public`
- `( --extended-key | --key-only )`

command-line/lib/Command/Key/Public.hs
KtorZ: More than which one is most common, I'd go for the "safest" from a privacy standpoint here. Since sharing your chain code allows other to create new (public) child keys from your public key, you may inadvertently gives access to an entire part of your sub-tree by sharing your extended public key. 

Plus, cardano-cli and other cardano-tools mostly deals with non-extended ed25519 keys only.

command-line/lib/Options/Applicative/Public.hs
paweljakubas: I somewhat just complied with the pattern we have for other things. So although I am slightly in favour with your opinion on this this would be outlier and may surprise anyone NEW looking at the code....

command-line/lib/Command/Key/Public.hs
paweljakubas: introduced `--with-chain-code` and `--without-chain-code` here https://github.com/input-output-hk/cardano-addresses/pull/87/commits/c2e3c7801beb75a5541c86a1a3d8e09f06d17459

command-line/lib/Command/Key/Public.hs
paweljakubas: thanks for suggestion! in the end solved this like that https://github.com/input-output-hk/cardano-addresses/pull/87/commits/4ad1a31aa59c09476b4ae27c6465833cc28425e3

README.md
paweljakubas: @KtorZ  fair point. I was thinking here in the following way. We use `prv`, `pub`, `xpub`, `pub` for keys...and for me mnemonic sequence is just text.... If that does not persuade you I will revert this in coming commit. @rvl do you have an opinion on that?

rvl: Yes and yes
1. It would be better if the `recovery-phrase generate` command were added back to provide context.
2. `.txt` is a good extension to describe the content of the file.

command-line/lib/Command/Key/Public.hs
rvl: if you use monadic rather than applicative then the ??? placeholder is not needed.

command-line/lib/Command/Key/Public.hs
```diff
@@ -40,13 +43,25 @@ mod liftCmd = command "public" $
...
+        <$> encodingOpt [humanReadablePart|???|]
+        <*> publicOpt
 
 run :: Cmd -> IO ()
-run Public{encoding} = do
+run Public{encoding, chainCode} = do
     xprv <- hGetXPrv stdin
     let xpub = toXPub xprv
-    hPutBytes stdout (xpubToBytes xpub) encoding
+    (bytes, encoding') <- case chainCode of
```
rvl, line 58: This line doesn't need to be an IO action. It could be a plain let or where clause.

I would be inclined to set up hrp in the CLI Parser, and use a function.

```haskell
let toBytes = case chainCode of
  WithChainCode -> xpubToBytes
  WithoutChainCode -> xpubPublicKey
hPutBytes stdout (toBytes xpub) encoding
```

command-line/lib/Options/Applicative/Public.hs
rvl: OK

command-line/lib/Command/Key/Public.hs
```diff
@@ -24,14 +24,17 @@ import Options.Applicative.Encoding
...
 import System.IO
     ( stdin, stdout )
 import System.IO.Extra
     ( hGetXPrv, hPutBytes )
 
 
-newtype Cmd = Public
+data Cmd = Public
     { encoding :: Encoding
+    , chainCode :: PublicType
```
rvl, line 37: Parser will look better if the order of constructor fields are swapped.

rvl (approved):
A couple style issues, but otherwise good.

KtorZ (approved):

PR #88, opened 2020-11-06
CIP-5 compliance

KtorZ:
- 1c061c5f86123111f2d4d95a1563bf6defcfaafb
  :round_pushpin: **expose all CIP-5 compliant prefixes in a dedicated module.**
  
- bfa84838d46fd123fd9a03e19454a2b6626547ef
  :round_pushpin: **remove encoding choice for all key commands and force bech32 with adequate prefixes.**
    It is still very easy to use raw hexadecimal binaries by using an
  intermediate tool like `bech32` to convert from hex to bech32. Yet,
  the human-readable part are crucial in making the implementation work
  well with multiple derivation levels, they also provide slightly
  stronger guarantee and a safer API.

  BREAKING-CHANGE: no more `--encoding` options available on any of the
                   `key` sub-commands. Commands now expect only bech32
                   inputs with specific prefixes (listed in case of
                   error).

  BREAKING-CHANGE: no more `--legacy` option available on the `key
                   child` command. The derivation scheme is now inferred
                   from the path and the key type.

  BREAKING-CHANGE: `key child` can now only derive from root to account,
                   root to address or account to address instead of any
                   arbitrary path.

- 42a60668526a98fbac2b491126ab9f64884aaa50
  :round_pushpin: **remove encoding choice for all address commands and force bech32 with adequate prefixes.**
    BREAKING-CHANGE: commands no longer accept arbitrary encoding for keys
                   but require that all keys are bech32 encoded and have
                   matching prefixes.

  BREAKING-CHANGE: credentials are now inferred from their bech32
                   prefixes (instead of `--from-key` or `--from-script`
                   options).

- 0aceac25eb49582c7559c9b115027e1850984495
  :round_pushpin: **Update README**
  
- 8cdf089de6d2e72b4653aac42252a3f98aea7a87
  :round_pushpin: **remove encoding choice for all address commands and force bech32 with adequate prefixes.**
    The commit also include some refactoring of the 'Script' module,
  namely:

  - it moves the script validation as part of the 'scriptFromString'
    function, so that we what returns this function is either an error
    or a structurally AND semantically valid script. This way, we don't
    need to perform validation further down at call sites. If a function
    receives a 'Script' data-type, it is expected to be valid.

  - it changes the script parser to only bech32-encoded keys, to be more
    consistent with the new overall approach.

- 9785bb27cf144a99c8df0020b00e8c5e9f2b9e4b
  :round_pushpin: **sprinkle some 'parallel' here and there in tests**
    - stack test cardano-addresses: Finished in 14.7219 seconds --> Finished in 3.3463 seconds
  - stack test cardano-addresses-cli: Finished in 23.1646 seconds --> Finished in 12.2827 seconds


command-line/lib/Command/Address/Payment.hs
```diff
@@ -65,28 +59,48 @@ mod liftCmd = command "payment" $
...
   where
-    hrp | networkTag == shelleyTestnet = [humanReadablePart|addr_test|]
-        | otherwise = [humanReadablePart|addr|]
+    addrHrp
+        | networkTag == shelleyTestnet = CIP5.addr_test
+        | otherwise = CIP5.addr
+
+    -- TODO: Also allow `XXX_vk` prefixes. We don't need the chain code to
+    -- construct a payment credential. This will however need some additional
+    -- abstraction over `xpubFromBytes` but I've done enough yake-shaving at
```
paweljakubas, line 82: we can use https://github.com/input-output-hk/cardano-addresses/blob/master/core/lib/Cardano/Address/Derivation.hs#L135 and then https://github.com/input-output-hk/cardano-addresses/blob/master/core/lib/Cardano/Address/Derivation.hs#L147 right?

command-line/lib/Command/Address/Payment.hs
```diff
@@ -65,28 +59,48 @@ mod liftCmd = command "payment" $
...
+    -- this stage, so leaving this as an item for later.
+    allowedPrefixes =
+        [ CIP5.addr_xvk
+        , CIP5.stake_xvk
+        , CIP5.script_xvk
```
paweljakubas, line 87: why do we allow `script_xvk` here? in credential from key either `addr_xvk` or `stake_xvk` can go, in credential from script we accept script hash which has `script` right?

command-line/lib/Command/Address/Reward.hs
```diff
@@ -70,24 +65,43 @@ mod liftCmd = command "stake" $
...
+        | otherwise = CIP5.stake
+
+    -- TODO: Also allow `XXX_vk` prefixes. We don't need the chain code to
+    -- construct a payment credential. This will however need some additional
+    -- abstraction over `xpubFromBytes` but I've done enough yake-shaving at
+    -- this stage, so leaving this as an item for later.
+    allowedPrefixes =
+        [ CIP5.addr_xvk
+        , CIP5.stake_xvk
+        , CIP5.script_xvk
```
paweljakubas, line 88: the same here... of course we can accept this and if detected then we need to hash it like provided in hashing before putting to DelegationFromScript

command-line/lib/Command/Key/Child.hs
```diff
@@ -56,34 +49,109 @@ mod liftCmd = command "child" $
...
+    --
+    -- There's no use-case at the moment for accessing intermediate paths such
+    -- as m / purpose' or m / purpose' / coin_type' so we do not expose them.
+    childHrpFor [_,_,_,2,_] hrp
+        | hrp == CIP5.root_xsk = pure CIP5.stake_xsk
+
+    childHrpFor [_,_,_,3,_] hrp
+        | hrp == CIP5.root_xsk = pure CIP5.script_xsk
+
+    childHrpFor [_,_,_,4,_] hrp
```
paweljakubas, line 121: we allow rule=3 and rule=4 for scripts ... Why rule=3 is not enough?

command-line/lib/Command/Address/Payment.hs
KtorZ: We cannot actually :/
Because xpubFromBytes expects 64 bytes of inputs, both a public key and the chain code. However, here, we would only have the public key.

We could use a "dummy" chain code filled with 0 bit, but that's quite hacky/ugly.

Since that wasn't supported before this PR anyway, I didn't really do anything to add it. But I think we should later.

command-line/lib/Command/Address/Payment.hs
KtorZ: Hmmm. That is correct. script_xvk should not be there because they would be part of scripts.

command-line/lib/Command/Key/Child.hs
```diff
@@ -56,34 +49,109 @@ mod liftCmd = command "child" $
...
+
+    childHrpFor [3,_] hrp
+        | hrp == CIP5.acct_xsk = pure CIP5.script_xsk
+        | hrp == CIP5.acct_xvk = pure CIP5.script_xvk
+
+    childHrpFor [4,_] hrp
+        | hrp == CIP5.acct_xsk = pure CIP5.script_xsk
+        | hrp == CIP5.acct_xvk = pure CIP5.script_xvk
+
+    childHrpFor [_,_] hrp
```
paweljakubas, line 142: thinking about this one if correct :thinking:

command-line/lib/Command/Address/Payment.hs
paweljakubas: `script_xvkh` keys are building script which in the end when hashed should have `script`

command-line/lib/Options/Applicative/Credential.hs
```diff
@@ -1,30 +1,48 @@
...
+delegationCredentialArg str = fromKey <|> fromScript <|> failure
   where
-    fromKeyFlag = flag' CredentialFromKey (long "from-key")
-    fromScriptFlag = flag' CredentialFromScript (long "from-script")
+    fromKey =
+        DelegationFromKey . liftXPub <$> xpubArg allowedPrefixes str
+      where
+        -- TODO: Allow non-extended keys here.
+        allowedPrefixes =
+            [ CIP5.stake_xvk
```
paweljakubas, line 39: don't we miss `addr_xvk` and `script` here? Ie., `stake_xvk` and `addr_xvk` should be prefixes for from key and `script` prefix should be for from script credential..

command-line/lib/Options/Applicative/Derivation.hs
```diff
@@ -191,42 +186,47 @@ derivationIndexToString ix_@(DerivationIndex ix)
...
-        "Couldn't detect input encoding? The key must be encoded as \
-        \base16, bech32 or base58."
-  where
-    toBytes = T.encodeUtf8 . T.pack
-
-xpubReader :: String -> Either String XPub
-xpubReader str = do
-    bytes <- encodingReader str
+xpubReader :: [HumanReadablePart] -> String -> Either String XPub
+xpubReader allowedPrefixes str = do
```
paweljakubas, line 193: only bech32 are legal...ok make sense to lower complexity of choices

command-line/lib/System/IO/Extra.hs
```diff
@@ -78,15 +83,31 @@ hGetBytes h = do
...
     decode from = either fail pure . from
 
+-- | Read some bytes encoded in Bech32, only allowing the given prefixes.
+hGetBech32 :: Handle -> [HumanReadablePart] -> IO (HumanReadablePart, ByteString)
+hGetBech32 h allowedPrefixes = do
+    raw <- B8.filter noNewline <$> B8.hGetContents h
+    (hrp, bytes) <- decode (fromBech32 markCharsRedAtIndices) raw
+    when (hrp `notElem` allowedPrefixes) $ fail
+        $ "Invalid human-readable prefix. Prefix ought to be one of: "
+        <> show (showHrp <$> allowedPrefixes)
```
paweljakubas, line 102: very nice response when not prefix is used :100:

paweljakubas (approved):
LGTM! Left some questions in places that were not so straightforward to me. 

Also, we should probably update README.md

README.md
```diff
@@ -127,43 +129,61 @@ pub16apaenn9ut6s40lcw3l8v68xawlrlq20z2966uzcx8jmv2q9uy7qssu5l3
...
-de5861cd05e99985b2c586ab383790c6600990809206f84e96eadaea
-$ cat signingKey2.xprv | cardano-address key public --with-chain-code | cardano-address key hash --base16 > verKey2.hash
-aca52d7d28ce353f4766e4e2c8cc2208c7113d794e776eafb8c07a80
-```
-Also notice the default hrp (in bech32) for the hashes of verification key - **xpub_hash**
-```
-$ cat signingKey1.xprv | cardano-address key public --with-chain-code | cardano-address key hash
-xpub_hash1mevxrng9axvctvk9s64nsduscesqnyyqjgr0sn5katdw5egajw2
+$ cat signingKey1.xprv \
+| cardano-address key public --without-chain-code \
```
piotr-iohk, line 159: It seems to be impossible to obtain verification key hash of the extended public key:

```
$ cat signingKey1.xprv \
| cardano-address key public --with-chain-code \
| cardano-address key hash \
| tee verKey1.hash

user error (Invalid human-readable prefix. Prefix ought to be one of: ["addr_vk","stake_vk","script_vk"])
```

Is that intentional to enforce compatibility with https://input-output-hk.github.io/cardano-wallet/api/edge/#operation/postAnyAddress?

README.md
```diff
@@ -189,8 +210,8 @@ script_hash15hx806zf0g8kcv399dpxf6fq4l98myqpvvzj2rltg465uz36435
 
   $ cat addr.prv \
   | cardano-address key public --with-chain-code \
-  | cardano-address address payment --from-key --network-tag testnet \
-  | cardano-address address delegation --from-script $(cardano-address script hash "$(cat script.txt)")
+  | cardano-address address payment --network-tag testnet \
+  | cardano-address address delegation $(cardano-address script hash "$(cat script.txt)")
```
piotr-iohk, line 214: This one actually does not work:

```
$ cat addr.prv \
  | cardano-address key public --with-chain-code \
  | cardano-address address payment --network-tag testnet \
  | cardano-address address delegation $(cardano-address script hash "$(cat script.txt)")
  
Invalid human-readable prefix. Prefix ought to be one of: ["stake_xvk"]

Usage: cardano-address address delegation (XPUB | SCRIPT HASH)
  Create a delegation address
```

I suppose the problem is in the `cardano-address address delegation` which doesn't want to accept script hash, although the message suggests `Usage: cardano-address address delegation (XPUB | SCRIPT HASH)
`:

```
$ cardano-address script hash "$(cat script.txt)" | tee script.hash
script1qd46g66dq98dh8h2qh947wq3qajf3plmlnkkedn2yexxke2lahd


$ cardano-address address delegation $(cat script.hash)
Invalid human-readable prefix. Prefix ought to be one of: ["stake_xvk"]

Usage: cardano-address address delegation (XPUB | SCRIPT HASH)
  Create a delegation address
```

piotr-iohk (commented):
I have pushed README update and left few comments.

command-line/lib/Command/Key/Child.hs
KtorZ: No, they translate to role=3 and role=4 from: https://hackmd.io/bdMljmoWR2yu4fqeAC7FCA

command-line/lib/Command/Key/Child.hs
KtorZ: Remember that Byron-legacy wallets have only 2 path segments.

command-line/lib/Options/Applicative/Credential.hs
KtorZ: The script is captured by the `fromScript` parser, which uses `scriptHashArg` from `Options.Applicative.Script`. As for `addr_xvk`, it's not allowed here because it's a delegation credential, so we expect either a stake key or a script.

command-line/lib/Cardano/Codec/Bech32/Prefixes.hs
```diff
@@ -0,0 +1,108 @@
...
+script_vk = [humanReadablePart|script_vk|]
+
+script_sk :: HumanReadablePart
+script_sk = [humanReadablePart|script_sk|]
+
+script_xvk :: HumanReadablePart
+script_xvk = [humanReadablePart|script_xvk|]
+
+script_xsk :: HumanReadablePart
+script_xsk = [humanReadablePart|script_xsk|]
```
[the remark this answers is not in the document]
KtorZ: see: https://github.com/cardano-foundation/CIPs/pull/31

README.md
KtorZ: It is intentional, to make it clear that the hash is about the public key itself, not the public key + chain code. We could accept both, and simply discard the chain code in the case of the extended one.

README.md
KtorZ: This one took me a while :o ... Apparently, we cannot combine multiple argument parsers with an alternative, they do not backtrack. So if the first one fails, the entire chain fails. That's a pity: 

https://github.com/pcapriotti/optparse-applicative/issues/166

KtorZ: Also, I didn't caught this earlier because there was actually no test scenario for constructing a delegation address from a script. I've added one :+1:

KtorZ:
I just noticed I forgot to remove the `Encoding` option on the `script hash` command too. Also, the script parser there seems to still expect base16 strings, which is not right.

PR #89, opened 2020-11-12
remove Jormungandr modules and related code.

KtorZ:
This is no longer applicable to this project which is about Cardano.


rvl (approved):
I grepped and couldn't find any references.
It would also need a ChangeLog entry.
:+1:

KtorZ:
>  It would also need a ChangeLog entry.

@rvl I'll add that (as well as all the breaking changes from #88) separately upon releasing :+1:

PR #90, opened 2020-11-12
Release/3.0.0

KtorZ:


.coverage/cardano-addresses.overlay
```diff
@@ -20,10 +20,6 @@ module "Cardano.Address.Derivation" {
   tick "getIndex" [Record Accessor];
 }
 
-module "Cardano.Address.Errors" {
-  tick function "showsPrec" [Show instance];
-}
-
```
KtorZ, line 26: That module no longer exists.

.github/workflows/main.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
     - name: 🔨 Build & Test
       run: |
-        stack --no-terminal test --bench --haddock --no-haddock-deps --no-run-benchmarks --flag cardano-addresses:release --flag cardano-addresses-cli:release
+        stack --no-terminal test --no-run-tests --bench --haddock --no-haddock-deps --no-run-benchmarks --flag cardano-addresses:release --flag cardano-addresses-cli:release
```
KtorZ, line 43: We are already running the tests right after as part of the coverage calculation yet, we still want to build the doc and with the proper "release" flag, but no need to run the tests twice.

piotr-iohk: ok, I was just about to ask...

piotr-iohk (approved):
lgtm

PR #92, opened 2020-11-13
fix missing pattern-match in the key hash command.

KtorZ:
Just caught this while testing something else and got bitten by the "impossible: pattern-match not coverage all cases.". Obviously, test cases for this very particular thing were missing so I can only blame it on myself :upside_down_face:


paweljakubas (approved):
looks fine

PR #94, opened 2020-11-17
Fix cabal package versions and bump to 3.1.1

rvl:
### Jira ticket

ADP-552

### Overview

Fixes problems found during the cardano-wallet v2020-11-17 release.

1. Fixes the `cardano-address --version` output, and bump to 3.1.1.
2. The `GITREV` environment variable can be used at build time to set the git version info, in case this program is not being built from a git clone.
3. Add a CI check to ensure that cabal files are up-to-date with respect to the hpack source.

### Comments

Note that the CI pipeline is currently broken because github removed a function - the failure is not caused by this PR. I have built it locally.


hasufell:
Any reason we use hpack at all?

jonathanknowles (approved):
LGTM!

PR #95, opened 2020-12-17
Extend script by timelocks

paweljakubas:
I have extended script to account for ActiveFromSlot and ActiveUntilSlot. I have updated all functions engaged, json instance, json generator, and script parser. also added script parser tests.

I have tested the correctness of resultant scriptHash in cardano-wallet tests -> https://github.com/input-output-hk/cardano-wallet/pull/2401

### DIFF
--- a/core/lib/Cardano/Address/Script.hs
+++ b/core/lib/Cardano/Address/Script.hs
@@ -95,8 +95,8 @@ data Script
     | RequireAllOf ![Script]
     | RequireAnyOf ![Script]
     | RequireSomeOf Word8 ![Script]
-    | ValidFromSlot Natural
-    | ValidUntilSlot Natural
+    | ActiveFromSlot Natural
+    | ActiveUntilSlot Natural
     deriving stock (Generic, Show, Eq)
 instance NFData Script
 
@@ -127,9 +127,9 @@ serializeScript script =
             , CBOR.encodeInt (fromInteger $ toInteger m)
             , encodeFoldable toCBOR contents
             ]
-        ValidFromSlot slotNum ->
+        ActiveFromSlot slotNum ->
             encodeMultiscriptCtr 4 2 <> CBOR.encodeWord64 (fromInteger $ toInteger slotNum)
-        ValidUntilSlot slotNum ->
+        ActiveUntilSlot slotNum ->
             encodeMultiscriptCtr 5 2 <> CBOR.encodeWord64 (fromInteger $ toInteger slotNum)
 
     encodeMultiscriptCtr :: Word -> Word -> CBOR.Encoding
@@ -252,8 +252,8 @@ foldScript fn zero = \case
     RequireAllOf xs      -> foldMScripts xs
     RequireAnyOf xs      -> foldMScripts xs
     RequireSomeOf _ xs   -> foldMScripts xs
-    ValidFromSlot _      -> zero
-    ValidUntilSlot _     -> zero
+    ActiveFromSlot _     -> zero
+    ActiveUntilSlot _    -> zero
   where
     foldMScripts =
         runIdentity . foldM (\acc -> Identity . foldScript fn acc) zero
@@ -287,9 +287,9 @@ validateScript = \case
         when (hasDuplicate script) $ Left DuplicateSignatures
         traverse_ validateScript script
 
-    ValidFromSlot _ -> pure ()
+    ActiveFromSlot _ -> pure ()
 
-    ValidUntilSlot _ -> pure ()
+    ActiveUntilSlot _ -> pure ()
   where
     hasDuplicate xs = do
         length sigs /= length (L.nub sigs)
@@ -357,7 +357,7 @@ prettyErrValidateScript = \case
 --          ]
 --}
 --{ "all" : [ "e09d36c79dec9bd1b3d9e152247701cd0bb860b5ebfd1de8abb6735a"
---          , {"valid_from": 120 }
+--          , {"active_from": 120 }
 --          ]
 --}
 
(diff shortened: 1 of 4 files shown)


core/lib/Cardano/Address/Script.hs
```diff
@@ -90,6 +95,8 @@ data Script
     | RequireAllOf ![Script]
     | RequireAnyOf ![Script]
     | RequireSomeOf Word8 ![Script]
+    | ValidFromSlot Natural
+    | ValidUntilSlot Natural
     deriving stock (Generic, Show, Eq)
```
KtorZ, line 100: Might be a good opportunity to add our lovely `Ord` instance to `Script` and `ScriptHash` and `KeyHash`don't you think :sweat_smile: ?

core/lib/Cardano/Address/Script.hs
```diff
@@ -120,6 +127,10 @@ serializeScript script =
             , CBOR.encodeInt (fromInteger $ toInteger m)
             , encodeFoldable toCBOR contents
             ]
+        ValidFromSlot slotNum ->
+            encodeMultiscriptCtr 4 2 <> CBOR.encodeWord64 (fromInteger $ toInteger slotNum)
```
KtorZ, line 131:
```suggestion
            encodeMultiscriptCtr 4 2 <> CBOR.encodeWord64 (fromIntegral slotNum)
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -120,6 +127,10 @@ serializeScript script =
...
+        ValidUntilSlot slotNum ->
+            encodeMultiscriptCtr 5 2 <> CBOR.encodeWord64 (fromInteger $ toInteger slotNum)
```
KtorZ, line 133:
```suggestion
            encodeMultiscriptCtr 5 2 <> CBOR.encodeWord64 (fromIntegral slotNum)
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -147,7 +158,7 @@ toScriptHash = ScriptHash . hashCredential . serializeScript
 --
 -- @since 3.0.0
 newtype ScriptHash = ScriptHash { unScriptHash :: ByteString }
-    deriving (Generic, Show, Eq)
+    deriving (Generic, Show, Ord, Eq)
```
KtorZ, line 161: Ah! You did :)

core/lib/Cardano/Address/Script.hs
```diff
@@ -255,6 +286,10 @@ validateScript = \case
         when (length script < fromIntegral m) $ Left ListTooSmall
         when (hasDuplicate script) $ Left DuplicateSignatures
         traverse_ validateScript script
+
+    ValidFromSlot _ -> pure ()
```
KtorZ, line 290: Is `0` an accepted value though :thinking: ? (I mean, purely from the protocol perspective, not in the context of a particular script).

core/lib/Cardano/Address/Script.hs
```diff
@@ -331,6 +370,10 @@ instance ToJSON Script where
         object ["any" .= fmap toJSON content]
     toJSON (RequireSomeOf count scripts) =
         object ["some" .= object ["at_least" .= count, "from" .= scripts]]
+    toJSON (ValidFromSlot slot) =
+        object ["valid_from" .= slot]
+    toJSON (ValidUntilSlot slot) =
+        object ["valid_until" .= slot]
```
KtorZ, line 376: On a second though, perhaps **active** captures this better than **valid**. Because, the script is valid from the moment it is created, but it may not be active yet / anymore. :thinking: ?

core/lib/Cardano/Address/Script/Parser.hs
```diff
@@ -72,6 +76,12 @@ scriptFromString str =
 --
 -- 5. Nested script are supported
 -- at_least 1 [3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3fe, all [3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3f1, 3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3f1]]
+-- 6. 1 signature required after slot number 120
+-- all [3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3fe, valid_from 120]
+-- 7. 1 signature required until slot number 150
+-- all [3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3fe, valid_until 150]
```
KtorZ, line 82: How to lock your token permanently 1-o-1 :smile:

KtorZ (approved):
Looks good. One reserve however about naming. I know I originally suggested this yesterday but after a night of sleep, doesn't "validXXX" sounds a bit weird?

core/lib/Cardano/Address/Script.hs
paweljakubas: not really -> https://github.com/input-output-hk/cardano-ledger-specs/blob/79055d5d482fb02c86f8b712eb4940f6e3c25a69/shelley-ma/impl/src/Cardano/Ledger/ShelleyMA/Timelocks.hs#L233

I plan to add someday time interval based validation - based on Allen interval algebra.

core/lib/Cardano/Address/Script/Parser.hs
paweljakubas: well easy ;-) :

core/lib/Cardano/Address/Script.hs
paweljakubas: yes, `active` is better -> adopted in https://github.com/input-output-hk/cardano-addresses/pull/95/commits/c5f72743677b1e153d5cd8b510f4d60e8f58f9ba

core/lib/Cardano/Address/Script.hs
paweljakubas: we are having `Natural` as a slot (to be least surprise as we use it also for `ChainPointer`). and we need to convert to Integer and then to Word64 like that to my best knowledge

core/lib/Cardano/Address/Script.hs
paweljakubas: we are having `Natural` as a slot (to be least surprise as we use it also for `ChainPointer`). and we need to convert to Integer and then to Word64 like that to my best knowledge

core/lib/Cardano/Address/Script.hs
paweljakubas: introduced `foldScript`, added `Ord` to `ScriptHash` and `KeyHash` to avoid things that were needed when adding script pool to  SeqState

paweljakubas:
added also golden tests in https://github.com/input-output-hk/cardano-addresses/pull/95/commits/c5f72743677b1e153d5cd8b510f4d60e8f58f9ba

PR #99, opened 2021-01-07
Introduce script template

paweljakubas:
ADP-637

Parametrizing Script
Adding Cosigner type
Adding ScriptTemplate type
Adding ToJSON and FromJSON instances of ScriptTemplate
Better error handling

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,14 +131,23 @@ script1dnt66jufkgx8rqxypxtz0hcrxs6hhayuj9cqh0eca82lcpwawd4
 </details>
 
 <details>
+  <summary>How to construct a multisig script hash with timelocks</summary>
+
+```console
+$  cardano-address script hash "all [$(cat script.1.xvk), $(cat script.2.xvk), active_from 100, active_until 120]"
+```
+</details>
+
+
+<details>
   <summary>How to validate a script</summary>
 
 ```console
-$  cardano-address script validate --required  "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk), $(cat script.2.xvk)]"
+$  cardano-address script validate "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk), $(cat script.2.xvk)]"
 Validated.
 
 $  cardano-address script validate --recommended  "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk), $(cat script.2.xvk)]"
-Not validated: The list inside a script has duplicate keys.
+Not validated: The list inside a script has duplicate keys (which is not recommended)..
 ```
 </details>
 
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ script_xvk1mg7xae48d7z4nntd35tey0jmclxaavwmk3kw2lkkt07p3s3x3yy45805manx2kj2neg40
 
 <details>
   <summary>How to generate a payment address from a payment key (<strong>payment.addr</strong>)</summary>
-  
+
 ```console
 $ cardano-address address payment --network-tag testnet < addr.xvk > payment.addr
 addr_test1vp2fg770ddmqxxduasjsas39l5wwvwa04nj8ud95fde7f7guscp6v
@@ -89,7 +89,7 @@ addr_test1vp2fg770ddmqxxduasjsas39l5wwvwa04nj8ud95fde7f7guscp6v
 
 <details>
   <summary>How to generate a delegated payment address from a stake key (<strong>payment-delegated.addr</strong>)</summary>
-  
+
 ```console
 $ cardano-address address delegation $(cat stake.xvk) < payment.addr > payment-delegated.addr
 addr_test1qp2fg770ddmqxxduasjsas39l5wwvwa04nj8ud95fde7f70k6tew7wrnx0s4465nx05ajz890g44z0kx6a3gsnms4c4qq8ve0n
@@ -98,7 +98,7 @@ addr_test1qp2fg770ddmqxxduasjsas39l5wwvwa04nj8ud95fde7f70k6tew7wrnx0s4465nx05ajz
 
 <details>
   <summary>How to generate a stake address from a stake key (<strong>stake.addr</strong>)</summary>
-  
+
 ```console
 $ cardano-address address stake --network-tag testnet < stake.xvk > stake.addr
 stake_test1urmd9uh08pen8c26a2fn86weprjh52638mrdwc5gfac2u2s25zpat%
@@ -118,21 +118,33 @@ script1qzzzlvn435jzdpm9dz5sk5helh6u2n5wa7g49m03sk4lzxhsxgt
 This script requires the signature from both signing keys corresponding to `script.1.xvk` and `script.2.xvk` in order to be valid. Similarly, we could require only one of the two signatures:
 
 ```console
-$ cardano-address script hash "any [$(cat script.1.xvk), $(cat script.2.xvk)]" 
+$ cardano-address script hash "any [$(cat script.1.xvk), $(cat script.2.xvk)]"
 script19raudr366cluzcwjxu67v7w50dazvxc0xtyfjy99wvmd593squy
 ```
 
 which is equivalent to:
 
 ```console
-$ cardano-address script hash "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk)]" 
+$ cardano-address script hash "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk)]"
 script1dnt66jufkgx8rqxypxtz0hcrxs6hhayuj9cqh0eca82lcpwawd4
 ```
 </details>
 
 <details>
+  <summary>How to validate a script</summary>
+
+```console
+$  cardano-address script validate --required  "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk), $(cat script.2.xvk)]"
+Validated.
+
+$  cardano-address script validate --recommended  "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk), $(cat script.2.xvk)]"
+Not validated: The list inside a script has duplicate keys.
+```
+</details>
+
+<details>
   <summary>How to generate a payment script address from a script hash (<strong>script.addr</strong>)</summary>
-  
+
 ```console
 $ cardano-address address payment --network-tag testnet < script.hash > script.addr
 addr_test1wqqggtajwkxjgf58v452jz6jl87lt32w3mhez5hd7xz6hugp80tta
--- a/command-line/lib/Command/Script/Validation.hs
+++ b/command-line/lib/Command/Script/Validation.hs
@@ -1,0 +1,81 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+module Command.Script.Validation
+    ( Cmd
+    , mod
+    , run
+
+    ) where
+
+import Prelude hiding
+    ( mod )
+
+import Cardano.Address.Script
+    ( KeyHash
+    , Script (..)
+    , TxValidity (..)
+    , ValidationLevel (..)
+    , prettyErrValidateScript
+    , validateScript
+    )
+import Data.Word
+    ( Word64 )
+import Options.Applicative
+    ( CommandFields
+    , Mod
+    , command
+    , footerDoc
+    , header
+    , helper
+    , info
+    , optional
+    , progDesc
+    )
+import Options.Applicative.Help.Pretty
+    ( bold, indent, string, vsep )
+import Options.Applicative.Script
+    ( levelOpt, scriptArg, txValidOpt )
+import System.IO
+    ( stdout )
+import System.IO.Extra
+    ( hPutString, progName )
+
+data Cmd = Cmd
+    { script :: Script KeyHash
+    , validationLevel :: ValidationLevel
+    , txValidFrom :: Maybe Word64
+    , txValidTo :: Maybe Word64
+    } deriving (Show)
+
+mod :: (Cmd -> parent) -> Mod CommandFields parent
+mod liftCmd = command "hash" $
+    info (helper <*> fmap liftCmd parser) $ mempty
+        <> progDesc "Validate a script"
+        <> header "Provide either required or recommended validation of a script given validation interval."
+        <> footerDoc (Just $ vsep
+            [ string "The script is taken as argument."
+            , string ""
+            , string "Example:"
+            , indent 2 $ bold $ string $ progName<>" script validate 'all "
+            , indent 4 $ bold $ string "[ 3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f3fe"
+            , indent 4 $ bold $ string ", 3c07030e36bfffe67e2e2ec09e5293d384637cd2f004356ef320f333"
+            , indent 4 $ bold $ string "]'"
+            , indent 2 $ string "a015ae61075e25c3d9250bdcbc35c6557272127927ecf2a2d716e29f"
+            ])
+  where
+    parser = Cmd
+        <$> scriptArg
+        <*> levelOpt
+        <*> optional (txValidOpt "tx-valid-from")
+        <*> optional (txValidOpt "tx-valid-to")
+
+run :: Cmd -> IO ()
+run Cmd{script,validationLevel,txValidFrom,txValidTo} = do
+    let txValidity = TxValidity (fromIntegral <$> txValidFrom) (fromIntegral <$> txValidTo)
+    case validateScript validationLevel txValidity script of
+        Left err -> hPutString stdout $ "Not validated: " <> prettyErrValidateScript err
+        Right _ -> hPutString stdout "Validated."
--- a/command-line/lib/Command/Script/Validation.hs
+++ b/command-line/lib/Command/Script/Validation.hs
@@ -79,7 +79,7 @@ mod liftCmd = command "validate" $
             , indent 2 $ string "Validated."
             , string ""
             , indent 2 $ bold $ string $ progName<>" script validate --recommended 'all []'"
-            , indent 2 $ string "Not validated: The list inside a script is empty."
+            , indent 2 $ string "Not validated: The list inside a script is empty or only contains timelocks (which is not recommended)."
             , string ""
             , indent 2 $ bold $ string $ progName<>" script validate 'at_least 1 [active_from 11, active_until 16]'"
             , indent 2 $ string "Validated."
@@ -94,7 +94,7 @@ mod liftCmd = command "validate" $
             , indent 4 $ bold $ string "[ script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms"
             , indent 4 $ bold $ string ", script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms"
             , indent 4 $ bold $ string "]'"
-            , indent 2 $ string "Not validated: The list inside a script has duplicate keys."
+            , indent 2 $ string "Not validated: The list inside a script has duplicate keys (which is not recommended)."
             ])
   where
     parser = Cmd
--- a/command-line/lib/Command/Script/Validation.hs
+++ b/command-line/lib/Command/Script/Validation.hs
@@ -62,8 +62,7 @@ mod liftCmd = command "validate" $
             , string " - check if there are no duplicated verification keys in a given level"
             , string " - check if 'at_least' coeffcient is positive"
             , string " - check if 'all', 'any' are non-empty and `'at_least' has no less elements in the list than the coeffcient after timelocks are filtered out. "
-            , string "The validation of the script does not take into account transaction validity. We assume that the wallet will take care of this upon sending"
-            , string "transaction with a given script."
+            , string "The validation of the script does not take into account transaction validity. We assume that the user will take care of this."
             , string ""
             , string "Example:"
             , indent 2 $ bold $ string $ progName<>" script validate 'all"
@@ -103,6 +102,6 @@ mod liftCmd = command "validate" $
 
 run :: Cmd -> IO ()
 run Cmd{script,validationLevel} =
-    case validateScript (fromMaybe RequiredValidation validationLevel) Nothing script of
+    case validateScript (fromMaybe RequiredValidation validationLevel) script of
         Left err -> hPutString stderr $ "Not validated: " <> prettyErrValidateScript err
         Right _ -> hPutString stdout "Validated."
--- a/command-line/lib/Command/Script/Validation.hs
+++ b/command-line/lib/Command/Script/Validation.hs
@@ -40,7 +40,7 @@ import Options.Applicative.Help.Pretty
 import Options.Applicative.Script
     ( levelOpt, scriptArg, txValidOpt )
 import System.IO
-    ( stdout )
+    ( stderr, stdout )
 import System.IO.Extra
     ( hPutString, progName )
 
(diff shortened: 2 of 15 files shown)


core/lib/Cardano/Address/Script.hs
```diff
@@ -504,14 +509,25 @@ instance FromJSON Cosigner where
...
+instance ToJSON XPub where
+    toJSON = String . T.decodeUtf8 . encode EBase16 . xpubToBytes
+
+instance FromJSON XPub where
+    parseJSON = withText "XPub" $ \txt ->
+        case fromBase16 (T.encodeUtf8 txt) of
+            Left err -> fail err
+            Right hex -> case xpubFromBytes hex of
+                Nothing -> fail "Extended public key cannot be retrieved from a given hex bytestring"
+                Just validXPub -> pure validXPub
```
KtorZ, line 571: I'd suggest making those simple functions :`encodeXPub :: XPub -> Aeson.Value` and `parseXPub :: Aeson.Value -> Aeson.Parser XPub` instead to avoid the orphan instance.

core/lib/Cardano/Address/Script.hs
```diff
@@ -147,10 +165,27 @@ serializeScript script =
...
+
+-- | Represents the script template that show the structure of the script and determines
+-- the expected place of verification keys corresponding to given cosigners.
+--
+-- @since 3.2.0
+data ScriptTemplate = ScriptTemplate
+    { cosigners :: Map Cosigner XPub
+    , template :: Script Cosigner
+    } deriving (Generic, Show, Eq)
+instance NFData ScriptTemplate
```
KtorZ, line 183: :+1:

core/lib/Cardano/Address/Script.hs
```diff
@@ -266,35 +301,75 @@ foldScript fn zero = \case
...
+    let validateKeyHash (KeyHash bytes) =
+            when (BS.length bytes /= credentialHashSize) $ Left WrongKeyHash
+    validateScript' validateKeyHash script
+
+validateScript'
+    :: Eq elem
+    => (elem -> Either ErrValidateScript ())
+    -> Script elem
+    -> Either ErrValidateScript ()
+validateScript' validateRequireSignatureOf = \case
```
KtorZ, line 315: I generally try to avoid `'` as much as possible when naming functions as there are often better alternative. Something like: `validateAbstractScript` or `validateScriptWith` for instance.

core/lib/Cardano/Address/Script.hs
```diff
@@ -266,35 +301,75 @@ foldScript fn zero = \case
...
+                ActiveFromSlot _ -> True
+                ActiveUntilSlot _ -> True
+                _ -> False
+        in case filter hasTimelocks xs of
+            [] -> False
+            [ActiveFromSlot s1, ActiveUntilSlot s2] -> s2 <= s1
+            [ActiveUntilSlot s2, ActiveFromSlot s1] -> s2 <= s1
+            [ActiveFromSlot _] -> False
+            [ActiveUntilSlot _] -> False
+            _ -> True
```
KtorZ, line 357: Arguably, `[ActiveFromSlot t1, ActiveFromSlot t2]` is va valid script (although one of the two clauses is redundant). I don't think we should do any particular validation with regards to scripts yet. It gives a false sense of safeness whereas there are still plenty of ways to shoot oneself in the foot.

core/lib/Cardano/Address/Script.hs
```diff
@@ -434,9 +497,126 @@ instance FromJSON Script where
...
+    ActiveFromSlot <$> o .: "active_from"
+
+parseActiveUntil
+    :: Value
+    -> Parser (Script elem)
+parseActiveUntil = withObject "Script ActiveUntil" $ \o ->
+    ActiveUntilSlot <$> o .: "active_until"
+
+instance ToJSON Cosigner where
+    toJSON (Cosigner ix) = object ["cosigner" .= toJSON ix]
```
KtorZ, line 556: It might be nicer to serialize cosigners as strings like `cosigner#1`, `cosigner#2` etc.. for it then allows to serialize the cosigner map as an object too by defining a `ToJSONKey Cosigner` instance.

core/lib/Cardano/Address/Script.hs
paweljakubas: done, see state as in https://github.com/input-output-hk/cardano-addresses/pull/99/commits/070e6bf1e09840eb7f88e782baed7eeee56dcabe

core/lib/Cardano/Address/Script.hs
paweljakubas: done, see state as in https://github.com/input-output-hk/cardano-addresses/pull/99/commits/070e6bf1e09840eb7f88e782baed7eeee56dcabe

core/lib/Cardano/Address/Script.hs
paweljakubas: I agree, one can construct any ScriptTemplate value. but if one wants to validate it and somewhat get feedback on what are reasonable things to use then can refer to `validateScriptTemplate`.  I do think using several ActiveFromSlot in a given nesting should be discouraged as it can give rise to problems one day...

core/lib/Cardano/Address/Script.hs
paweljakubas: good idea.done, see state as in https://github.com/input-output-hk/cardano-addresses/pull/99/commits/070e6bf1e09840eb7f88e782baed7eeee56dcabe

core/lib/Cardano/Address/Script.hs
```diff
@@ -82,30 +95,34 @@ import qualified Codec.Binary.Bech32 as Bech32
...
+data Script (elem :: *)
+    = RequireSignatureOf !elem
+    | RequireAllOf ![Script elem]
+    | RequireAnyOf ![Script elem]
+    | RequireSomeOf Word8 ![Script elem]
     | ActiveFromSlot Natural
     | ActiveUntilSlot Natural
     deriving stock (Generic, Show, Eq)
-instance NFData Script
+instance NFData elem => NFData (Script elem)
```
KtorZ, line 118: :+1:

core/lib/Cardano/Address/Script.hs
```diff
@@ -266,35 +300,75 @@ foldScript fn zero = \case
...
+    => (elem -> Either ErrValidateScript ())
+    -> Script elem
+    -> Either ErrValidateScript ()
+validateScriptWith validateRequireSignatureOf = \case
+    RequireSignatureOf element ->
+        validateRequireSignatureOf element
 
     RequireAllOf script -> do
-        when (L.null script) $ Left EmptyList
+        when (L.null (omitTimelocks script)) $ Left EmptyList
```
KtorZ, line 319: Why the `omitTimeLocks` here :thinking: Surely, we can create a script with only timelocks, no ?

core/lib/Cardano/Address/Script.hs
```diff
@@ -266,35 +300,75 @@ foldScript fn zero = \case
...
+-- | Validate a 'ScriptTemplate', semantically
+--
+-- @since 3.2.0
+validateScriptTemplate :: ScriptTemplate -> Either ErrValidateScript ()
+validateScriptTemplate (ScriptTemplate cosigners' script) = do
+    when (Map.size cosigners' == 0) $ Left NoCosigner
+    when (L.length (L.nub $ Map.elems cosigners') /= Map.size cosigners') $
+        Left DuplicateXPubs
+    let validateCosigner cosigner =
+            when (cosigner `notElem` (Map.keys cosigners')) $ Left UnknownCosigner
```
KtorZ, line 370: I think we need to check inclusion both ways: 

- All cosigners referred to in the cosigners map are in the script
- All cosigners mentioned in the script are also in the cosigners map

So that we enforce a clear 1:1 mapping.

core/lib/Cardano/Address/Script.hs
```diff
@@ -328,7 +406,15 @@ prettyErrValidateScript = \case
     Malformed ->
         "Parsing of the script failed. The script should be composed of nested \
         \lists, and the verification keys should be either encoded as bech32."
-
+    InvalidTimelocks ->
+        "The list inside a script must contain at most two timelock conditions \
+        \and they cannot be contradictory."
+    DuplicateXPubs ->
+        "Teh cosigners in a script template must assume an unique extended public key."
```
KtorZ, line 413: Typos: `Teh` -> `The`, `an unique` -> `a unique`

core/lib/Cardano/Address/Script.hs
```diff
@@ -328,7 +406,15 @@ prettyErrValidateScript = \case
...
+    UnknownCosigner ->
+        "The script must use a cosigner present in a script template."
+    NoCosigner ->
+        "The script template must have at least one cosigner defined."
```
KtorZ, line 417: It feels somewhat wrong to have these two last cases as part of the same error type for script validation. May I suggest to have instead 2 sum-types, one with errors that concerns `Script KeyHash` and one with others that concerns `Script Cosigner`? 

It's maybe not easy to re-conciliate the implementation with `validateScriptWith` ... we would need to pull off type families and GADTs likely... which may not be worth it for something _that_ simple.

core/test/Cardano/Address/ScriptSpec.hs
```diff
@@ -296,28 +331,158 @@ spec = do
...
+        let cosigner2 = RequireSignatureOf (Cosigner 2)
+        let cosigner3 = RequireSignatureOf (Cosigner 3)
+
+        it "no cosigners in script template" $ do
+            let scriptTemplate = ScriptTemplate Map.empty (RequireAllOf [])
+            validateScriptTemplate scriptTemplate `shouldBe` (Left NoCosigner)
+
+        it "illegal cosigner in script template" $ do
+            let scriptTemplate = ScriptTemplate cosigners' (RequireSignatureOf (Cosigner 4))
+            validateScriptTemplate scriptTemplate `shouldBe` (Left UnknownCosigner)
```
KtorZ, line 364: I'd like to see the other way around too, an extra cosigner in the map that is not mentioned in the script ;)

core/lib/Cardano/Address/Script.hs
```diff
@@ -306,6 +380,10 @@ data ErrValidateScript
     | DuplicateSignatures
     | WrongKeyHash
     | Malformed
+    | InvalidTimelocks
+    | DuplicateXPubs
```
KtorZ, line 384: Couldn't this case be solved by modelling the ScriptTemplate differently? For instance, use a `Set (Cosigner, XPub)`. In practice, the set will be reasonably bounded (and we could/should very much enforce that) so searching in the set for a particular index would be fast. Plus, it automatically enforces uniqueness of elements.

core/test/Cardano/Address/ScriptSpec.hs
```diff
@@ -296,28 +331,158 @@ spec = do
...
+                    ])
+            validateScriptTemplate scriptTemplate `shouldBe` (Left InvalidTimelocks)
+
+        it "too high m in RequireSomeOf when timelocks" $ do
+            let scriptTemplate = ScriptTemplate cosigners' (RequireSomeOf 3 [cosigner0, cosigner1, ActiveFromSlot 21, ActiveUntilSlot 30])
+            validateScriptTemplate scriptTemplate `shouldBe` (Left ListTooSmall)
+
+        it "no content in RequireAnyOf when timelocks" $ do
+            let scriptTemplate = ScriptTemplate cosigners' (RequireAnyOf [ActiveFromSlot 21, ActiveUntilSlot 30])
+            validateScriptTemplate scriptTemplate `shouldBe` (Left EmptyList)
```
KtorZ, line 450: I don't see why this is an invalid script? Or, is it something actually forbidden by the ledger? I'd suggest to revise the error type then if it is because `EmptyList` is a bit confusing, the list isn't empty but rather lacks of required key hashes (if they are indeed required?). I don't recall this however :thinking: ?

core/test/Cardano/Address/ScriptSpec.hs
```diff
@@ -296,28 +331,158 @@ spec = do
...
+    describe "can perform roundtrip JSON serialization & deserialization - Script KeyHash" $
+        it "fromJSON . toJSON === pure" $ property (prop_jsonRoundtrip @(Script KeyHash))
+    describe "can perform roundtrip JSON serialization & deserialization - Script KeyHash validated" $
+        it "fromJSON . toJSON === pure" $ property (prop_jsonRoundtripWithValidation validateScript)
+    describe "can perform roundtrip JSON serialization & deserialization - Script Cosigner" $
+        it "fromJSON . toJSON === pure" $ property (prop_jsonRoundtrip @(Script Cosigner))
+    describe "can perform roundtrip JSON serialization & deserialization - ScriptTemplate" $
+        it "fromJSON . toJSON === pure" $ property (prop_jsonRoundtrip @ScriptTemplate)
+    describe "can perform roundtrip JSON serialization & deserialization - ScriptTemplate validated" $
+        it "fromJSON . toJSON === pure" $ property (prop_jsonRoundtripWithValidation validateScriptTemplate)
```
KtorZ, line 465: :+1:

core/lib/Cardano/Address/Script.hs
paweljakubas: indeed, from the ledger point of view it can be acceptable, depends on slots used, where it is used, how it related to tx validity...made distinction between required (as understood by ledger) and recommended validation

core/lib/Cardano/Address/Script.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: fixed

core/lib/Cardano/Address/Script.hs
paweljakubas: not we have separate script and separate script template errors.  one of script template error is when script is wrong, but validation of key hashes and cosigners is decoupled. please refer to code

core/test/Cardano/Address/ScriptSpec.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: yes, I used set operation to resolve this. please refer to HEAD

core/test/Cardano/Address/ScriptSpec.hs
paweljakubas: not ledger (required) and recommended validation is present. There are numerous tests that show what is required, what is recommended (aka I found reasonable)

command-line/lib/Command/Script/Validation.hs
```diff
@@ -0,0 +1,94 @@
...
+    } deriving (Show)
+
+mod :: (Cmd -> parent) -> Mod CommandFields parent
+mod liftCmd = command "validate" $
+    info (helper <*> fmap liftCmd parser) $ mempty
+        <> progDesc "Validate a script"
+        <> header "Choose a required or recommended validation of a script for a given transaction validation interval."
+        <> footerDoc (Just $ vsep
+            [ string "The script is taken as argument. To have required validation pass '--required'."
+            , string "To have recommended validation pass '--recommended'. Recommended validation includes also"
```
KtorZ, line 61: It'd be good to explain what `recommended` adds compared to `required`.

command-line/lib/Command/Script/Validation.hs
```diff
@@ -0,0 +1,94 @@
...
+            , string ""
+            , indent 2 $ bold $ string $ progName<>" script validate --recommended --tx-valid-from 10 --tx-valid-to 15 'at_least 1 [active_from 11, active_until 16]'"
+            , indent 2 $ string "Not validated: At least's number must not be larger than the non-timelock elements in the list."
+            ])
+  where
+    parser = Cmd
+        <$> scriptArg
+        <*> levelOpt
+        <*> optional (txValidOpt "tx-valid-from")
+        <*> optional (txValidOpt "tx-valid-to")
```
KtorZ, line 87: I am not sure to understand what the `tx-valid-from` and `tx-valid-to` have to do here :thinking: ?

command-line/test/Command/ScriptSpec.hs
```diff
@@ -55,27 +58,85 @@ spec = do
...
+        specScriptHashProper "script1uan07g0z30cfpvg7gxp5fmvjmv6ylaqa84msw8twyt5777hf7gf"
             [iii|any [ #{verKeyH1}, #{verKeyH2}, #{verKeyH1}]|]
 
         specScriptInvalid Malformed
             [iii|script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxjq8egs9|]
 
+        specScriptInvalid Malformed
+            [iii|any [ #{verKeyH1}, #{verKeyH2}, active_from a]|]
+
+    describeCmd [ "script", "validate"] $ do
```
KtorZ, line 88: What about splitting this module in two: `test/Command/Script/HashSpec` and `test/Command/Script/ValidateSpec` to follow the same pattern as other test files.

core/lib/Cardano/Address/Script.hs
```diff
@@ -215,6 +255,29 @@ keyHashFromText txt = do
...
+data TxValidity = TxValidity (Maybe Natural) (Maybe Natural)
+    deriving (Show, Eq, Generic)
+instance NFData TxValidity
+
+-- Validation level. Required level does basic check that will make sure the script
+-- is accepted in ledger. Recommended level collects a number of checks that will
+-- warn about dangerous, unwise and redundant things present in the script.
+--
+-- @since 3.2.0
+data ValidationLevel = RequiredValidation | RecommendedValidation
```
KtorZ, line 277: :+1:

core/lib/Cardano/Address/Script.hs
```diff
@@ -258,77 +321,201 @@ foldScript fn zero = \case
...
 --
 -- @since 3.0.0
 prettyErrValidateScript
     :: ErrValidateScript
     -> String
 prettyErrValidateScript = \case
+    LedgerIncompatible ->
+        "The script is ill-formed and is not going to be accepted by ledger."
     EmptyList ->
         "The list inside a script is empty."
```
KtorZ, line 487: Note that with the `omitTimelock` above, this message may be a bit confusing and should be reworked. "The list inside a script is empty or only contains timelocks (which is not recommended)."

core/lib/Cardano/Address/Script.hs
```diff
@@ -258,77 +321,201 @@ foldScript fn zero = \case
...
+    -> TxValidity
+    -> Script KeyHash
+    -> Either ErrValidateScript ()
+validateScript level interval script = do
+    let validateKeyHash (KeyHash bytes) =
+            (BS.length bytes == credentialHashSize)
+    let allSigs = foldScript (:) [] script
+    unless (L.all validateKeyHash allSigs) $ Left WrongKeyHash
+    unless (requiredValidation interval script)
+        $ Left LedgerIncompatible
```
KtorZ, line 342: Hmm. We loose quite a bit of information with this opaque error type. Wouldn't it be better to have `requiredValidation` return a proper error rather than a boolean? I'd suggest making a new sum-type `ErrRequiredValidation` to cover the various errors that can arise from `requiredValidation` and create a new branch in `ErrValidateScript` to hold those possible errors.

core/lib/Cardano/Address/Script.hs
```diff
@@ -258,77 +321,201 @@ foldScript fn zero = \case
...
+    when (level == RecommendedValidation ) $
+        recommendedValidation script
+
+requiredValidation
+    :: Eq elem
+    => TxValidity
```
KtorZ, line 348: I am not quite sure about this argument. It is very unlikely that scripts are going to be created with this knowledge in mind. Scripts are typically created well before transactions. So, this is a check we could perform upon submitting a transaction but not before.

core/lib/Cardano/Address/Script.hs
```diff
@@ -258,77 +321,201 @@ foldScript fn zero = \case
...
     :: ErrValidateScript
     -> String
 prettyErrValidateScript = \case
+    LedgerIncompatible ->
+        "The script is ill-formed and is not going to be accepted by ledger."
     EmptyList ->
         "The list inside a script is empty."
     MZero ->
-        "At least must be at least 1."
+        "At least's number must be not smaller than 1."
```
KtorZ, line 489: suggestion: number -> coefficient maybe ?

command-line/test/Command/Script/ValidationSpec.hs
```diff
@@ -0,0 +1,102 @@
...
+
+        specScriptNotValidated DuplicateSignatures RecommendedValidation
+            [iii|any [ #{verKeyH1}, #{verKeyH2}, #{verKeyH2}]|]
+
+        specScriptValidated RequiredValidation
+            [iii|at_least 0 [ #{verKeyH1}, #{verKeyH2} ]|]
+
+        specScriptNotValidated MZero RecommendedValidation
+            [iii|at_least 0 [ #{verKeyH1}, #{verKeyH2} ]|]
+
```
piotr-iohk, line 75: Some interesting cases I stumbled upon. I thought they will all validate as all are "well-formed"... So I guess `--required` does check some logic after all. Now I'm a bit more confused as of what is the difference between `--required` and `--recommended` :sweat_smile:  

```
cardano-address script validate --required "any []"
Not validated: The script is ill-formed and is not going to be accepted by ledger.

cardano-address script validate --required "at_least 1 []"
Not validated: The script is ill-formed and is not going to be accepted by ledger.

cardano-address script validate --required "at_least 4 [$(cat script.1.xvk), $(cat script.2.xvk)]"
Not validated: The script is ill-formed and is not going to be accepted by ledger.

cardano-address script validate --required "at_least 0 []"
Validated.

```

command-line/lib/Command/Script/Validation.hs
paweljakubas: added explanation plus make --required default and whole option optional, see HEAD

command-line/lib/Command/Script/Validation.hs
paweljakubas: yes, after reconsideration I opted for removing this from command-line. tx validity and hence proper choice of ttl, etc. and validating this should be on consumer side (eg. wallet). So I left possibility to use validation with TxValidity in function, but not in command line. see HEAD

command-line/test/Command/ScriptSpec.hs
paweljakubas: done, see HEAD

core/lib/Cardano/Address/Script.hs
paweljakubas: refactored errors, also added (I hope) better messaging. see HEAD

core/lib/Cardano/Address/Script.hs
paweljakubas: refactored errors to reflect what is required (be compliant with ledger, not malformed in syntax sense, have legal verification keys), what is recommended. see HEAD

core/lib/Cardano/Address/Script.hs
paweljakubas: made it optional. as I think it should stay here, this possibility to check it for client of this library (eg. wallet). But in command line I use Nothing. see HEAD

README.md
```diff
@@ -118,21 +118,33 @@ script1qzzzlvn435jzdpm9dz5sk5helh6u2n5wa7g49m03sk4lzxhsxgt
...
 ```console
-$ cardano-address script hash "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk)]" 
+$ cardano-address script hash "at_least 1 [$(cat script.1.xvk), $(cat script.2.xvk)]"
 script1dnt66jufkgx8rqxypxtz0hcrxs6hhayuj9cqh0eca82lcpwawd4
 ```
 </details>
 
+<details>
+  <summary>How to validate a script</summary>
+
```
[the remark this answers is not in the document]
paweljakubas: done, see head in my branch

command-line/test/Command/Script/ValidationSpec.hs
paweljakubas: that is what ledger accepts. So in summary in required we have 1. ledger compliant validation, 2. syntax validation 3 checking if verification keys are valid (by looking at length and prefix of bech32. And we have recommended validation. I hope now it is clear. And it is reflected in Error hierarchy. see head of this PR

piotr-iohk (approved):
to me looks good.

KtorZ (approved):
:+1:

PR #105, opened 2021-02-16
cleaning and fixing

paweljakubas:
This PR does two things
a. fixes how recommended validation treats timelocks in particular when we have:
(a) ALL [active_until s1, active s2,..]
 <0---------s1)
                             <s2,........+inf)
this is contradictory for each point in interval and should be not recommended as it could result in permanent funds locking
(b) ALL [active_until s1, active s2,..]
if s1==s2 then it is unlocked for s1
(c) ALL [active_until s1, active s2,..]
if s1 > s2 then it is unlocked for <s2,s1)

Also I introduced recommendations for any, here
(a) ANY [active_until s1, active s2,..]
 <0---------s1)
                             <s2,........+inf)
makes perfect sense. it is locked in <s1,s2) interval, for the rest unlocked
(b) if s1>s2 it is always satisfied so those locks are redundant

B. Updated documentation to reflect current state


core/lib/Cardano/Address/Script.hs
```diff
@@ -362,11 +362,13 @@ recommendedValidation = \case
         when (L.null (omitTimelocks script)) $ Left EmptyList
         when (hasDuplicate script) $ Left DuplicateSignatures
         when (redundantTimelocks script) $ Left RedundantTimelocks
+        when (timelockTrap script) $ Left TimelockTrap
```
rvl, line 365: What does TimelockTrap mean?
Is it basically that the interval ends before it starts?

rvl (approved):
More tests seems good.
The error constructor names don't make much sense to me though.

rvl:
What's the "recommended" validation for?

core/lib/Cardano/Address/Script.hs
paweljakubas: it it the situation when we have script ALL with timelocks. if we have for example all [active_until 10, active_after 20, something else] then in order to have this section satisfied all elements are to be satisfied. But it is not possible to be BOTH before slot=10 and after slot=20 at the same time! So we DO NOT RECOMMEND it in validation as it means the script will lock spending from funds or staking from them. Although ledger would accept this script, and hence user can, on his own responsibility use required validation, and pursue with this script

paweljakubas:
Required validation checks if script will be accepted by ledger. Recommended validation is additional validation on top of required validation that checks more things, like the one involving funds trapping above.

core/test/Cardano/Address/ScriptSpec.hs
```diff
@@ -447,14 +447,59 @@ spec = do
...
+                                   , RequireAllOf [cosigner0, cosigner1, cosigner2 ,cosigner3, ActiveFromSlot 25, ActiveUntilSlot 21]
+                                   ]
+                    ])
+            validateScriptTemplate RecommendedValidation scriptTemplate `shouldBe` Left (WrongScript $ NotRecommended TimelockTrap)
+
+        it "valid timelocks when using any" $ do
+            let scriptTemplate = ScriptTemplate cosigners' (RequireSomeOf 1
+                    [ cosigner1
+                    , RequireAnyOf [ cosigner2
+                                   , RequireAnyOf [cosigner0, cosigner1, cosigner2 ,cosigner3, ActiveFromSlot 25, ActiveUntilSlot 21]
```
KtorZ, line 484: I don't get why this is considered valid ?

core/test/Cardano/Address/ScriptSpec.hs
```diff
@@ -447,14 +447,59 @@ spec = do
...
+                                   ]
+                    ])
+            validateScriptTemplate RecommendedValidation scriptTemplate `shouldBe` Right ()
+
+        it "invalid timelocks when using any 1" $ do
+            let scriptTemplate = ScriptTemplate cosigners' (RequireSomeOf 1
+                    [ cosigner1
+                    , RequireAnyOf [ cosigner2
+                                   , RequireAnyOf [cosigner0, cosigner1, cosigner2 ,cosigner3, ActiveFromSlot 21, ActiveUntilSlot 25]
```
KtorZ, line 493: And why this isn't? More exactly, I'd expect both this example and the above to be unsound.

core/test/Cardano/Address/ScriptSpec.hs
paweljakubas: well, to validate this part the following is needed:
(a) any time one of signatures of any cosigner : 0, 1, 2, 3

core/test/Cardano/Address/ScriptSpec.hs
paweljakubas: it is not recommended, because is is redundant as it is always valid. It says it should be ok either <0,25) V <21,+inf) which means in whole slot domain <0,+inf) so it is always true. so why to include it?

core/test/Cardano/Address/ScriptSpec.hs
KtorZ: Okay. My reasoning was that basically having scripts in the shape of `any [signature, timelock]` is generally unsound. When specifying a timelock, I think you almost always want to use `all` or `at_least n` with n >= 2.

PR #106, opened 2021-03-04
Add ghcjs to CI and fix build issues

hamishmack:
[Hydra jobset PR #106](https://hydra.iohk.io/jobset/Cardano/cardano-addresses-pr-106#tabs-jobs)

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -14,6 +14,13 @@ source-repository-package
     tag: 2547ad1e80aeabca2899951601079408becbc92c
     --sha256: 1p2kg2w02q5w1cvqzhfhqmxviy4xrzada3mmb096j2n6hfr20kri
 
+source-repository-package
+    type: git
+    location: https://github.com/hamishmack/foundation
+    tag: 421e8056fabf30ef2f5b01bb61c6880d0dfaa1c8
+    --sha256: 0cbsj3dyycykh0lcnsglrzzh898n2iydyw8f2nwyfvfnyx6ac2im
+    subdir: foundation
+
 -- Relax overly strict bounds in hjsonschema and hjsonpointer
 allow-newer:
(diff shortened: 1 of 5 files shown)


cabal.project
```diff
@@ -14,6 +14,13 @@ source-repository-package
     tag: 2547ad1e80aeabca2899951601079408becbc92c
     --sha256: 1p2kg2w02q5w1cvqzhfhqmxviy4xrzada3mmb096j2n6hfr20kri
 
+source-repository-package
+    type: git
+    location: https://github.com/hamishmack/foundation
```
rvl, line 19: Because this build fix is so small, could we do it as a patch instead of forking on github?

nix/sources.json
```diff
@@ -1,14 +1,14 @@
 {
     "haskell.nix": {
-        "branch": "master",
+        "branch": "hkm/emcc",
```
rvl, line 3: Fine to use a branch for now - I assume there will be a PR some time?

Does this mean it actually uses emcc to compile the cbits?

rvl (approved):
Awesome, thanks!

nix/sources.json
hamishmack: I'm still not sure I have it right.  Will check with @luite.  Once it is we can merge it and set this back to master.

cabal.project
hamishmack: I looked into this and it won't work as a patch.  We need `cabal configure` to see the patched version and haskell.nix can only patch the sources selected after `cabal configure` runs.

rvl: OK - fair enough!

nix/sources.json
rvl: OK, so #108.

PR #110, opened 2021-03-11
ADP-744 bech32 in inspectAddress

reactormonk:
This PR introduces the `ToBech32` / `FromBech32` classes to support bech32 in `inspectAddress`, and propagates these changes throughout the code.

Breaking change: `bech32` how has type `Text -> Either String a` instead of `Text -> Maybe a`.

For formatting, I ran `stylish-haskell -i -c .stylish-haskell.yaml core/**/*.hs command-line/**/*.hs` which reformatted the entire codebase :raised_eyebrow:


.gitignore
```diff
@@ -26,7 +26,8 @@ cabal.project.local~
 *.tix
 *.lock
 hie.yaml
+.vscode
```
KtorZ, line 29: Note that this has nothing to do with the project, whatever concerns your editor is none of our business 😅

You should consider using a global git ignore for your tooling / editor.

command-line/lib/Command/Address/Delegation.hs
```diff
@@ -77,7 +79,7 @@ run Cmd{credential} = do
         Left (ErrInvalidAddressType  msg) ->
             fail msg
         Right addr ->
-            B8.hPutStr stdout $ T.encodeUtf8 $ bech32With hrp addr
+           B8.hPutStr stdout $ T.encodeUtf8 $ bech32With hrp (unAddress addr)
```
KtorZ, line 82: Shouldn't this be using ToBech32 instead of unwrapping the Address?

command-line/lib/Command/Address/Pointer.hs
```diff
@@ -90,7 +92,7 @@ run Cmd{slotNum,transactionIndex,outputIndex} = do
         Left (ErrInvalidAddressType  msg) ->
             fail msg
         Right addr ->
-            B8.hPutStr stdout $ T.encodeUtf8 $ bech32With hrp addr
+            B8.hPutStr stdout $ T.encodeUtf8 $ bech32With hrp (unAddress addr)
```
KtorZ, line 95: Same here? (use ToBech32)

core/lib/Cardano/Address.hs
```diff
@@ -64,7 +61,10 @@ import GHC.Stack
 import Numeric.Natural
     ( Natural )
 
-import qualified Cardano.Codec.Bech32.Prefixes as CIP5
+import Cardano.Codec.Bech32
+    ( FromBech32 (..), ToBech32 (..), bech32With, fromBech32With )
+import Cardano.Codec.Bech32.Prefixes
+    ( addr, addr_vkh, stake_vkh )
```
KtorZ, line 67: Please use a qualified import for this one to give some context to various hrp and remain consistent with other usages of that module.

command-line/lib/Command/Address/Delegation.hs
reactormonk: It can contain two different prefixes, so no. I haven't yet though about how to handle prefix hierarchies (here both `addr` and `addr_test`).

reactormonk:
What would be the prefixes for these instances?

```
instance ToBech32 (WithoutChain (Shelley 'DelegationK XPrv)) where
    bech32 = undefined
instance ToBech32 (WithoutChain (Shelley 'PaymentK XPrv)) where
    bech32 = undefined
instance ToBech32 (WithoutChain (Shelley 'ScriptK XPrv)) where
    bech32 = undefined
```

.gitignore
reactormonk: Btw, I use Arch. ;-)

rvl:
Make sure you are using stylish-haskell-0.11.0.3.
Anything newer can't find the right config file in multi-package projects.
It ends up using the defaults.

rvl (commented):
The network tag (mainnet/testnet) is needed for encoding addresses in bech32 according to CIP5.
This value is the `network` variable in the `inspectAddress` function.
You will probably need to ditch the type classes I'm afraid.

core/lib/Cardano/Codec/Bech32.hs
```diff
@@ -0,0 +1,32 @@
...
+bech32With hrp =
+    T.decodeUtf8 . encode (EBech32 hrp)
+
+fromBech32With :: HumanReadablePart -> (ByteString -> b) -> Text -> Either String b
+fromBech32With hrn con text = do
+  (hrp, bs) <- E.fromBech32 (const id) $ T.encodeUtf8 text
+  if hrp == hrn then
+    pure $ con bs
+  else
+    Left $ "Human Readable Part should be " <> show (humanReadablePartToText hrn) <> " but is " <> show (humanReadablePartToText hrp)
```
jonathanknowles, line 32: Hi @reactormonk!

So we should probably mention that the Adrestia repositories are covered by a set of [Coding Standards](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards) that we try to apply to all new code in PRs.

For the section above, there are two issues to fix:

1. The code is indented with 2 spaces. However, we use 4 space indents in our repositories (except for the `where` keyword, which is indented with 2 spaces).
2. The line length is (much) longer than our suggested line limit: 80 chars ([source](https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters)).

Note that we do sometimes allow things to exceed the 80 char limit if it would be prohibitive to adhere to it, but in many cases, it's simple to fix, so we try to adhere to it when possible.

Here's a suggested code fix:

```suggestion
fromBech32With :: HumanReadablePart -> (ByteString -> b) -> Text -> Either String b
fromBech32With hrn con text = do
    (hrp, bs) <- E.fromBech32 (const id) $ T.encodeUtf8 text
    if hrp == hrn then
        pure $ con bs
    else
        Left $ mconcat
            [ "Human Readable Part should be "
            , show (humanReadablePartToText hrn)
            , " but is "
            , show (humanReadablePartToText hrp)
            ]
```

core/lib/Cardano/Address.hs
```diff
@@ -237,3 +220,15 @@ invariantNetworkTag limit (NetworkTag num)
       ++ show num
       ++ ", but expected to be less than "
       ++ show limit
+
+newtype AddressVerificationKeyHash = AddressVerificationKeyHash { unAddressVerificationKeyHash :: ByteString }
```
jonathanknowles, line 224:
```suggestion
newtype AddressVerificationKeyHash =
    AddressVerificationKeyHash { unAddressVerificationKeyHash :: ByteString }
```

https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters

core/lib/Cardano/Address.hs
```diff
@@ -237,3 +220,15 @@ invariantNetworkTag limit (NetworkTag num)
...
+instance ToBech32 AddressVerificationKeyHash where
+    bech32 (AddressVerificationKeyHash kh) = bech32With CIP5.addr_vkh kh
+instance FromBech32 AddressVerificationKeyHash where
+    fromBech32 = fromBech32With CIP5.addr_vkh AddressVerificationKeyHash
+
+newtype StakeVerificationKeyHash = StakeVerificationKeyHash { unStakeVerificationKeyHash :: ByteString }
```
jonathanknowles, line 230:
```suggestion
newtype StakeVerificationKeyHash =
    StakeVerificationKeyHash { unStakeVerificationKeyHash :: ByteString }
```

https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#limit-line-length-to-80-characters

jonathanknowles (commented):
Hi @reactormonk 

I've added a few requests / suggestions in comments.

command-line/lib/Command/Address/Delegation.hs
KtorZ: I do agree with Rodney here, if we can't use the abstraction in places where it would make sense then it is of little use :no_mouth: 

However, there's enough information already in an address to know which prefix should be used. The first byte of an address does contain network information. 

See: https://github.com/input-output-hk/cardano-ledger-specs/blob/master/shelley-ma/shelley-ma-test/cddl-files/shelley-ma.cddl#L71-L79

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -147,12 +155,20 @@ import GHC.Generics
...
+    ( FromBech32 (..), ToBech32, bech32, bech32With, fromBech32With )
+import qualified Cardano.Codec.Bech32.Prefixes as CIP5
 import qualified Data.Aeson as Json
 import qualified Data.ByteArray as BA
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Lazy as BL
+import Data.Coerce
+    ( coerce )
+import Data.Either.Extra
+    ( maybeToEither )
```
KtorZ, line 171: On grouping imports, have a look at: https://github.com/input-output-hk/adrestia/wiki/Coding-Standards#stylish-haskell-is-used-to-format-grouped-imports--language-pragmas

Do not interleave qualified with explicit imports :+1:

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -190,6 +206,88 @@ newtype Shelley (depth :: Depth) key = Shelley
...
+instance FromBech32 (Shelley 'ScriptK XPrv) where
+    fromBech32 b = do
+        dec <- fromBech32With CIP5.script_xsk id b
+        Shelley <$> maybeToEither "Invalid Key Length" (xprvFromBytes dec)
+instance FromBech32 (Shelley 'ScriptK XPub) where
+    fromBech32 b = do
+        dec <- fromBech32With CIP5.script_xvk id b
+        Shelley <$> maybeToEither "Invalid Key Length" (xpubFromBytes dec)
+
+newtype WithoutChain key = WithoutChain key
```
KtorZ, line 274: I'd suggest: `WithoutChainCode`, words are cheap, so let's use proper qualified when we can :+1:

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -190,6 +206,88 @@ newtype Shelley (depth :: Depth) key = Shelley
...
+
+instance ToBech32 (WithoutChain (Shelley 'RootK XPub)) where
+    bech32 k = bech32With CIP5.script_vk $ xpubPublicKey $ coerce k
```
KtorZ, line 277: Minor suggestion: that's a typical case where I find that pointfree style read actually better.

```hs
    bech32 = bech32With CIP5.script_vk . xpubPublicKey . coerce
```

core/lib/Cardano/Codec/Bech32.hs
KtorZ, line 32: On a different note, avoid showing back an invalid input. If `hrp` does not validate, we don't really know what it is. Can be a string of 324239 random characters and we don't want that printed back to the console or in an error message. The error should simply tell what the expected value is, and that it isn't. Users know what they've entered.

core/test/Cardano/Address/Script/ParserSpec.hs
```diff
@@ -8,34 +8,33 @@ module Cardano.Address.Script.ParserSpec
...
 spec = do
     let verKeyH1 = "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyreluzt36ms" :: Text
-    let (Right (_, bytes1)) = fromBech32 (const id) $ T.encodeUtf8 verKeyH1
+    let (Right bytes1) = fromBech32 verKeyH1
     let verKeyH2 = "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyrenxv223vj" :: Text
-    let (Right (_, bytes2)) = fromBech32 (const id) $ T.encodeUtf8 verKeyH2
+    let (Right bytes2) = fromBech32 verKeyH2
     let verKeyH3 = "script_vkh18srsxr3khll7vl3w9mqfu55n6wzxxlxj7qzr2mhnyre5g2sfvk2" :: Text
-    let (Right (_, bytes3)) = fromBech32 (const id) $ T.encodeUtf8 verKeyH3
+    let (Right bytes3) = fromBech32 verKeyH3
```
KtorZ, line 37: I guess they're no longer `bytes` so the name is a bit misleading now.

core/test/Cardano/Address/Style/ShelleySpec.hs
```diff
@@ -166,53 +163,36 @@ spec = do
...
+                    "018a4d111f71a79169c50bcbc27e1e20b6e13e87ff8f33edc3cab419d408b2d658668c2e341ee5bda4477b63c5aca7ec7ae4e3d196163556a4"
             }
 
     describe "Test vectors" $ do
         testVectors TestVector
             { rootXPrv =
-                    "root_xprv1nq49elkxhet9d03y0uqxp5fa802jgqauef3878ve3r9xvkg8\
-                    \c4qqnq98raswxtrwlfw7txa7m49d6x84092l0lsv5s603dkakj8vwgsz23\
-                    \drjsarjmxdev8mnrllmxykxh7exsrtqn64d6j5qdqp9deqpqx70t25"
+                    "root_xsk1nq49elkxhet9d03y0uqxp5fa802jgqauef3878ve3r9xvkg8c4qqnq98raswxtrwlfw7txa7m49d6x84092l0lsv5s603dkakj8vwgsz23drjsarjmxdev8mnrllmxykxh7exsrtqn64d6j5qdqp9deqpqlu52ew"
```
KtorZ, line 179: These changes seems both unnecessary and wrong? 

- The code was formatted to fit nicely on 80 columns.  
- Nothing done in this PR should affect these tests. If the test are failing, there's likely something wrong with the new implementation and it's the implementation that should be fixed, not the tests :grimacing:

KtorZ: I guess the change comes from the prefix changes from `root_xprv` to `root_xsk`, which changes the entire encoding (and similarly for acct and test like). 
In such case, it'd be _okay_ to adjust the tests indeed. However, unless strictly necessary, I'd avoid touching them as much as possible. 

This could be done in the test itself, by changing the prefix to what's expected by the test (with a little note explaining the rationale). This way, we do still test the new implementation, but keep tests the same (which ease our minds because it doesn't look like we've broken everything).

KtorZ (commented):
It'd be nice to see roundtrip `bech32 . fromBech32` for every instances of `ToBech32` / `FromBech32`. 

Also, on a small naming note, it'd be better to have either `toBech32 / ToBech32` or `bech32 / Bech32` instead of `bech32 / ToBech32`. Consistency is better.

KtorZ:
> @reactormonk 
> ```hs
> What would be the prefixes for these instances?
>
> instance ToBech32 (WithoutChain (Shelley 'DelegationK XPrv)) where
>    bech32 = undefined
> instance ToBech32 (WithoutChain (Shelley 'PaymentK XPrv)) where
>   bech32 = undefined
> instance ToBech32 (WithoutChain (Shelley 'ScriptK XPrv)) where
>    bech32 = undefined
> ```

- `WithoutChain (Shelley 'DelegationK XPrv)` corresponds to `stake_sk`
- `WithoutChain (Shelley 'PaymentK XPrv)` corresponds to `addr_sk`
- `WithoutChain (Shelley 'ScriptK XPrv)` corresponds to `script_sk`

core/lib/Cardano/Address/Style/Shelley.hs
reactormonk: I was hoping I could just not look at imports and stylish-haskell would take care of it.

core/lib/Cardano/Address/Style/Shelley.hs
reactormonk: I've always regretted point-free style whenever I went back to edit some code, I've been it avoiding it since. Could use it if it's more in style.

core/lib/Cardano/Address/Style/Shelley.hs
jonathanknowles: > I was hoping I could just not look at imports and stylish-haskell would take care of it.

It will ... **_mostly_**. 😅 

But AFAIK, qualified imports must be placed at the end manually.

It's an annoyance, I agree.

core/test/Cardano/Address/Style/ShelleySpec.hs
reactormonk: > The code was formatted to fit nicely on 80 columns.

so I could easily `sed` with copy/paste. It was about the limit of my frustration tolerance. I could replace them with a golden test.

> prefix changes

Yeah, they change the prefixes, and hence the checksum changes.

> This could be done in the test itself, by changing the prefix to what's expected by the test (with a little note explaining the rationale).

That sounds like trouble down the line, so I'd prefer avoiding it. I'd go with a golden test here so you can compare the output from the old one with the new one side-by-side. Or I could write myself a little macro to reformat the current code, seems to be mechanical enough.

command-line/lib/Command/Address/Delegation.hs
Anviking: > However, there's enough information already in an address to know which prefix should be used. The first byte of an address does contain network information.

That _would_ have been nice, but the stake keys don't contain the network id, only normal addresses

KtorZ: Stake addresses do contain a network id. And prefixes is only different between testnet and mainnet for stake addresses, not stake keys.

KtorZ: `stake` /= `stake_vk`

core/lib/Cardano/Address/Style/Shelley.hs
rvl: stylish-haskell will do most of it -- it you're using the version which works (see my other comment). But you need to group the imports yourself.

command-line/lib/Command/Address/Delegation.hs
Anviking: > Stake addresses do contain a network id

Ah, ok!

core/lib/Cardano/Address/Style/Shelley.hs
rvl: I love point-free style when used for simple unary functions. Just read the pipeline from right to left.

.gitignore
```diff
@@ -29,4 +29,4 @@ hie.yaml
 
 ### Nix ###
 result*
-.stack-to-nix.cache
+.stack-to-nix.cache
```
rvl, line 32: Your editor appears to be ignoring `.editorconfig`.
But at least it's not saving files with windows line endings.

core/lib/Cardano/Address.hs
```diff
@@ -77,6 +76,10 @@ newtype Address = Address
     { unAddress :: ByteString
     } deriving stock (Generic, Show, Eq, Ord)
 instance NFData Address
+instance ToBech32 Address where
+    bech32 (Address k) = bech32With CIP5.addr k
```
rvl, line 80: My thought was that it would be easiest to just not use a ToBech32 typeclass, and make a function instead.

But @KtorZ had a neat suggestion about using a typeclass for determining the CIP5 hrp. I quote him here. Choose what you think is best.

> - I like simple functions as well over type classes for encoding. I prefer a `addressToBech32` and `scriptToBech32` than a `ToBech32` type-class.
> - On the other hand I also like consistency, and consistency is easier to achieve with a type-class.
> - Beside consistency, another reason for using a type-class here would be to avoid repeating / duplicating CIP's prefixes everywhere. We could possibly get away with a single type class `class CIP5 t where hrpFor :: Proxy t -> HumanReadablePart` which gives you what prefix to use for a particular data-type.
> - Overall, a type-class to connect the CIP's prefix to a data-type makes sense to me here because we really are encoding a same capability of data-types that depends on the data-types themselves.

rvl (changes requested):
Hi @reactormonk, do you need any more info to address the review comments?

.gitignore
reactormonk: Plugin installed.

core/lib/Cardano/Address.hs
reactormonk: I've tried a few things with the `CIP5` typeclass only, without a `ToBech32`. Also, for the `Address` type, the `CIP5` human-readable part also depends on the content, so a `Proxy` isn't enough there. But then the API goes weird, where you have to pass things twice. A bunch of functions instead of a `ToBech32` class would also be possible, but then you'd have about 5-6 of these, with the `AddressVerificationKeyHash` newtypes. It gets _really_ interesting with the `Shelley` newtype, as you'll need separate functions for `XPub` and `XPrv`. I think I rest my case that a `ToBech32` typeclass is the best abstraction level we'll get.

reactormonk:
I did an overhaul.

- `CIP5` typeclass won't work, see https://github.com/input-output-hk/cardano-addresses/pull/110#discussion_r598734028
- deleted `FromBech32`, because it doesn't serve a purpose. Most inputs accept multiple types of addresses, so this doesn't look like the right abstraction.

rvl (changes requested):
Since we now know that `ToBech32` can't work, `FromBech32` is redundant, and `CIP5 t` is too hard, I would expect the diff to be dramatically smaller.

core/lib/Cardano/Address.hs
```diff
@@ -78,6 +87,29 @@ newtype Address = Address
     } deriving stock (Generic, Show, Eq, Ord)
 instance NFData Address
 
+instance ToBech32 Address where
+    bech32 addr = bech32With (addressHrp addr) $ unAddress addr
```
rvl, line 91: This is a little dodgy because `Address` may not be a shelley address.
It would not be good to expose this as API.
`addressHrp` is good and could be used, but only within context where the type of address is known to be Shelley.

rvl (commented):
Thanks @reactormonk - it's looking better now.
Could we put back all the bech32 functions where they originally were, and not have the `Cardano.Codec.Bech32` module?

core/lib/Cardano/Address.hs
reactormonk: Any address that isn't Shelley shouldn't be bech32 anyway.

rvl: Yes, so that's one of the reasons why `instance ToBech32 Address` is dodgy.

rvl (approved):
Thanks @reactormonk. I applied some review cleanups and removed some API breaks which weren't necessary for this feature.

PR #112, opened 2021-03-12
Remove Enum instances that don't work on 32-bit targets

rvl:
### Jira ticket

ADP-699

### Overview

These are @luite's modifications for 32-bit targets, with some review.

### Comment

[Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-addresses-pr-112#tabs-jobs)

### DIFF
--- a/core/lib/Cardano/Address/Derivation.hs
+++ b/core/lib/Cardano/Address/Derivation.hs
@@ -6,7 +6,6 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE ViewPatterns #-}
 
 {-# OPTIONS_HADDOCK prune #-}
 
@@ -18,7 +17,8 @@ module Cardano.Address.Derivation
     -- * Key Derivation
     -- ** Types
       Index
-    , Indexed (..)
+    , indexToWord32
+    , IndexFromWord32 (..)
     , wholeDomainIndex
     , coerceWholeDomainIndex
     , nextIndex
@@ -59,6 +59,7 @@ module Cardano.Address.Derivation
     , generateNew
     , hashCredential
     , credentialHashSize
+    , unsafeMkIndex
     ------------------
     ) where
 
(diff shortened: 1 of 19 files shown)


paweljakubas (approved):
lgtm and surprised that so many ripple effects this change induced;-)

some hint from coding standards to apply : 
```
core/test/Test/Arbitrary.hs:122:17-62: Warning: Use mapMaybe
Found:
  catMaybes . map fromWord32 . shrink . toWord32
Perhaps:
  mapMaybe fromWord32 . shrink . toWord32
```

core/lib/Cardano/Address/Derivation.hs
```diff
@@ -333,29 +341,55 @@ instance Bounded (Index 'WholeDomain depth) where
...
-        | Index (fromIntegral ix) > maxBound @(Index 'WholeDomain _) =
-            error "Index@WholeDomain.toEnum: bad argument"
-        | otherwise =
-            Index (fromIntegral ix)
+-- | Serialization of 'Index' to and from 'Word32'.
+--
+-- @since 3.3.0
+class Indexed a where
+    fromWord32 :: Word32 -> Maybe a
+    toWord32 :: a -> Word32
```
jonathanknowles, line 349: Hi @rvl!

This is just a suggestion, but perhaps it would be better if the class name `Indexed` and function names `{fromWord32,toWord32}` could more clearly relate to one another.

Some thoughts:

- This class deals exclusively with things that are convertible to and from `Word32`.
- It's also a replacement for `Enum` that works with `Word32` instead of `Int`.

Perhaps call it `Enum32`?

Then we'd have:

```hs
class Enum32 a where
    toEnum32 :: Word32 -> Maybe a
    fromEnum32 :: a -> Word32
```
(Note that the `to` and `from` functions are going in the same direction as the `Enum` class.)

rvl: Thanks for the ideas. I would like to distance this class from `Enum`, which has features and characteristics that we don't need for address path indices. I was thinking SerialiseAsWord32. But actually, now I think this class should be distilled down to just the function `indexFromWord32` (more like what Luite originally did...).

PR #114, opened 2021-03-25
CI and build fixes

rvl:
### Issue

ADP-744

### Description

- Fix the hpack check in CI which wasn't working before
- Fix the hspec-golden config. Only the "golden" files should be checked in. The "actual" files are generated locally, and should be ignored.
- Various clean ups of stack.yaml and cabal.project.

### Comments

- About the issue of ScriptSpec failing depending on who's running it. I suspect that the `allow-newer: true` option in `stack.yaml` allows this to happen.


reactormonk (approved):

reactormonk:
> About the issue of ScriptSpec failing depending on who's running it. I suspect that the allow-newer: true option in stack.yaml allows this to happen.

Do we have any version bounds in cabal files? I don't really see any, so I suspect the cabal index state dictates a different version than the one from stackage, so I suspect this could be solved via https://hackage.haskell.org/package/stack2cabal or using nix & cabal.

PR #118, opened 2021-04-07
Fix CI

rvl:
### Issue

ADP-860

### Description

1. It turns out that tests were being run in GitHub actions CI, but test failures would not cause CI to fail. This change makes stack execute tests after building them. Tests are run twice now, but oh well.
2. Some attempts at tightening version bounds to avoid test result discrepancies between developers and CI.
3. Add the checks attrset to `default.nix` for local testing.

### Comments

[Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-addresses-pr-118#tabs-jobs)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
 
     - name: 🔨 Build & Test
       run: |
-        stack --no-terminal test --no-run-tests --bench --haddock --no-haddock-deps --no-run-benchmarks --flag cardano-addresses:release --flag cardano-addresses-cli:release
+        stack --no-terminal test --bench --haddock --no-haddock-deps --no-run-benchmarks --flag cardano-addresses:release --flag cardano-addresses-cli:release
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
     - name: 🔨 Build & Test
       run: |
-        stack --no-terminal test --no-run-tests --bench --haddock --no-haddock-deps --no-run-benchmarks --flag cardano-addresses:release --flag cardano-addresses-cli:release
```
piotr-iohk, line 46: I believe the rationale for adding `--no-run-tests` was this https://github.com/input-output-hk/cardano-addresses/pull/90/files#r522033013

_(but perhaps it's a bit confusing from the first glance)_

rvl: It's very confusing.

rvl: I looked at the master branch CI build logs and it turns out the tests do actually run.
They fail, but the result is still green.
At least with this change, if tests fail then CI turns red.

KtorZ (approved):

PR #119, opened 2021-04-09
CLI: fix script command example

rvl:
While helping someone with cardano-adresses, I noticed that a new required CLI argument is missing from the examples.


piotr-iohk (approved):
lgtm

piotr-iohk:
Perhaps it would be beneficial to add e2e tests here similar to cardano-wallet's, that would just run through examples from a CLI...

PR #120, opened 2021-04-09
Add a shell.nix with ghcjs support

hamishmack:
Use `js-unknown-ghcjs-cabal` in the shell to build with ghcjs.

Uses https://github.com/input-output-hk/haskell.nix/pull/1089

The JS dependencies (those that we generate with emscripten) are moved into a separate package `cardano-addresses-jsbits`.  This package is not included in the `shellFor` `packages` list, so it will be built using nix.  That way the version of the `-jsbits` package used in the shell is the same as the one used when building with nix-build.

I had hoped to make it a sub library component rather than a package, but cabal would not pick up the sub library from the hackage database provided by haskell.nix (not sure if the problem was in haskell.nix or cabal).

Unfortunately `js-unknown-ghcjs-cabal run exe:cardano-address` does not work because it does not know to use `node`, but it is easy enough to run it manually.

```
[nix-shell:~/iohk/cardano-addresses]$ js-unknown-ghcjs-cabal run exe:cardano-address
Up to date
/home/hamish/iohk/cardano-addresses/dist-newstyle/build/wasm32-none/ghcjs-8.6.5/cardano-addresses-cli-3.3.0/x/cardano-address/build/cardano-address/cardano-address: createProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

[nix-shell:~/iohk/cardano-addresses]$ node dist-newstyle/build/wasm32-none/ghcjs-8.6.5/cardano-addresses-cli-3.3.0/x/cardano-address/build/cardano-address/cardano-address.jsexe/all.js
Usage: all.js ((-v|--version) | COMMAND)
  Command-line for address and key manipulation in Cardano.

Available options:
  -h,--help                Show this help text
  -v,--version             Show the software current version and build revision.

Available commands:
  version                  Show the software current version and build revision.
  recovery-phrase          About recovery phrases
  key                      About public/private keys
  address                  About addresses
  script                   About scripts

💡 Need auto-completion?

  ↳ source <( all.js --bash-completion-script `which all.js`)

Or alternatively --fish-completion-script / --zsh-completion-script.
For a long-term solution, you may want to put this script in the relevant place. e.g.:

  ↳ /etc/bash_completion.d

```


rvl:
Very cool @hamishmack

rvl (commented):
This looks good, thanks @hamishmack. I added my own re-arrangements - could you please check that they are OK? When PR input-output-hk/haskell.nix#1089 is merged we can update niv and approve this. I'm also still waiting for `shell.nix` to evaluate before I can test it. :slightly_smiling_face:

hamishmack:
The remaining problems are:
* that the `js-sources` from the library do not seem to be linked into the final exes (they seem to make it into the library `.js_a`, but are not in the `.jsexe`.   I have included a work around of adding the `.js` file to the `ghc-options` for now.
* some tests are failing because the JSON outputs keys in a different order (I wonder if there is a hashing algorithm used differs depending on size of Int).

rvl:
It's sort of working for me now, but something's not quite right with using the crypto lib at runtime:
```
$ js-unknown-ghcjs-cabal --builddir=dist-ghcjs build all
$ node /home/rodney/iohk/cardano-addresses/dist-ghcjs/build/wasm32-none/ghcjs-8.6.5/cardano-addresses-cli-3.3.0/x/cardano-address/build/cardano-address/cardano-address.jsexe/all.js recovery-phrase generate
uncaught exception in Haskell main thread: ReferenceError: h$cryptonite_sha256_init is not defined
```

hamishmack:
> uncaught exception in Haskell main thread: ReferenceError: h$cryptonite_sha256_init is not defined

The `js-sources` are not making it into the final exe files.  It is odd because I thought that bit was working before, but I might have tested it when I still had `ghc-options` as a work around.  We might need @luite to figure out what is missing.

In the mean time if you add  `__trace cardano-crypto-js` to `nix/haskell.nix` and pass the result as `--ghc-options=/nix/store/...` to `js-unknown-ghcjs-cabal` (or add `ghc-options: /nix/store/...` to the `cardano-addresses-cli.cabal` file) it should work.

rvl:
Thanks @hamishmack - that works! :tada:

Could it be because there are no jsbits in the registered cardano-addresses-jsbits package?
I was looking in the paths shown by:

```
js-unknown-ghcjs-ghc-pkg describe cardano-addresses-jsbits-3.3.0
```

Anyway, I made some more tweaks, and added build instructions for GHCJS.

Perhaps we could merge this PR now, then fix the limitations later.

rvl (approved):

hamishmack:
@rvl, I am happy for us to merge this, but I don't have access myself "The base branch restricts merging to authorized users.".  I'll show @luite the `js-sources` issue.

hamishmack:
This is WIP again.  The latest commit fixes GHCJS in the shell, but breaks the tests for GHC.  I'll fix that up now.

rvl:
Nice @hamishmack - that works for me. I re-organised so that it doesn't need CPP.

I was wondering whether the `addJsbitsDependency` function would only need to be invoked once from the library package somewhere, and that would be enough to link in the js-sources for all executables and tests?

hamishmack:
>  I re-organised so that it doesn't need CPP.

Cool. I tried to do it that way at first, but messed something up (I think I must have left out a dependency or something).

PR #122, opened 2021-04-09
Add tests against CLI examples printed by cardano-addresses

piotr-iohk:
**Motivation:** https://github.com/input-output-hk/cardano-addresses/pull/119#issuecomment-816455977

**Context:**
Every now and then we encounter that CLI example printed by `cardano-address` is out of date and needs to be fixed (like https://github.com/input-output-hk/cardano-addresses/pull/119). I thought it might be good idea to test those examples in kind of end-to-end fashion. This little test suite gets all the examples from the CLI and runs them making sure each command succeeds. 

**Output:**

```
$ rspec

CLI examples
  cardano-address address payment
  cardano-address address bootstrap
  cardano-address address stake
  cardano-address address delegation
  cardano-address address pointer
  cardano-address script
  cardano-address key

Finished in 0.15025 seconds (files took 0.22205 seconds to load)
7 examples, 0 failures
```

Clear message in case of error:

```
     Failure/Error: expect(cmd(ex_cmd)).to succeed
     
               Following command failed!!!
               --------------------------------------------
               cardano-address key inspect <<< $(cat acct.prv)
     
               StdErr:
               --------------------------------------------
               sh: 1: Syntax error: redirection unexpected
     
     
               StdOut:
               --------------------------------------------

```

**TODO:**
 - ~wire into GH actions~ https://github.com/input-output-hk/cardano-addresses/pull/122/checks?check_run_id=2308362484
 - ~check if it works on Windows~ - it doesn't. Examples are not tailored for Windows cmd (there's `cat`, `tee`, `<<<` which are not available on Win)
 - ~I'm having some problems with examples from `cardano-address key` subcommand. The examples use `<<<` which seems not to be supported by shell (I suppose only bash supports it) - try to fix it.~

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,11 @@ jobs:
     - name: 📥 Checkout repository
       uses: actions/checkout@v1
 
+    - name: 💎 Setup Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7.1
+
     - name: 🧰 Setup Stack
       uses: mstksg/setup-stack@v1
 
@@ -43,6 +48,7 @@ jobs:
 
     - name: 🔨 Build & Test
       run: |
+        echo "$GITHUB_WORKSPACE/bin" >> $GITHUB_PATH
         stack --no-terminal test --bench --haddock --no-haddock-deps --no-run-benchmarks --flag cardano-addresses:release --flag cardano-addresses-cli:release
         mkdir -p bin && cp $(stack path --local-install-root)/bin/* bin && chmod +x bin/*
         mkdir -p dist/haddock && mv $(stack path --local-install-root)/doc/* dist/haddock
@@ -52,6 +58,13 @@ jobs:
         DESTDIR=dist/coverage make report && DESTDIR=dist/coverage make badge
         echo "Checking for updated cabal files"
         git diff --exit-code -- '**/*.cabal' && echo OK
+
+    - name: 🧪 Test CLI examples
+      run: |
+        cd test/e2e
+        bundle install
+        rspec .
+        cd -
 
     - name: 📎 Upload Artifact
       uses: actions/upload-artifact@v1
--- a/nix/stack-shell.nix
+++ b/nix/stack-shell.nix
@@ -16,6 +16,7 @@ mkShell rec {
     stack
     nix
     pkgconfig
+    bundler
   ] ++ lib.optional (!stdenv.isDarwin) git;
 
   libs = [
--- a/test/e2e/.gitignore
+++ b/test/e2e/.gitignore
@@ -1,0 +1,5 @@
+*.rspec_status
+*.prv
+*.pub
+*.xprv
+*.hash
--- a/test/e2e/.gitignore
+++ b/test/e2e/.gitignore
@@ -1,5 +1,1 @@
 *.rspec_status
-*.prv
-*.pub
-*.xprv
-*.hash
--- a/test/e2e/.rspec
+++ b/test/e2e/.rspec
@@ -1,0 +1,3 @@
+--format documentation
+--color
+--require spec_helper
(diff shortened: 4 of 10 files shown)


rvl (changes requested):
This is a good idea, but I'm not keen on a DIY testing framework, when there are probably existing tools available which can do this tests.

This one seems like it would be good: https://github.com/fugue/goldplate

piotr-iohk:
> This is a good idea, but I'm not keen on a DIY testing framework, when there are probably existing tools available which can do this tests.

Ok, yeah my idea was to get examples from the `cardano-address` output and run them to check if they are OK, like from `cardano-address key`:

```
$ cardano-address key
....

Example:
  $ cardano-address recovery-phrase generate --size 15 \
    | cardano-address key from-recovery-phrase Shelley > root.prv
  
  $ cat root.prv \
    | cardano-address key child 1852H/1815H/0H \
    | tee acct.prv \
    | cardano-address key public --with-chain-code > acct.pub
  
  $ cardano-address key inspect <<< $(cat acct.prv)
  {
      "key_type": "private",
      "chain_code": "67bef6f80df02c7452e20e76ffb4bb57cae8aac2adf042b21a6b19e4f7b1f511",
      "extended_key": "90ead3efad7aacac242705ede323665387f49ed847bed025eb333708ccf6aa54403482a867daeb18f38c57d6cddd7e6fd6aed4a3209f7425a3d1c5d9987a9c5f"
  }
  
  $ cardano-address key inspect <<< $(cat acct.pub)
  {
      "key_type": "public",
      "chain_code": "67bef6f80df02c7452e20e76ffb4bb57cae8aac2adf042b21a6b19e4f7b1f511",
      "extended_key": "d306350ee88f51fb710252e27f0c40006c58e994761b383e02d400e2be59b3cc"
  }

```

I believe that https://github.com/fugue/goldplate and in frameworks alike you mostly define some tests in their own format, so even if we'd like to employ it some DYI would be required (that is for testing the examples)...

My goal was not to add e2e tests, I think they are pretty good already in https://github.com/input-output-hk/cardano-addresses/tree/master/command-line/test, but just explicitly test the examples, which were going out of date every once in a while.

rvl:
Oh, I see, it runs the example - good idea.

Could we have this test all in Hspec though? It would fit in quite well with the existing tests.

If the recovery phrase generate command is replaced with a fixed mnemonic, then we could also test expected values. e.g.
```
cardano-address address bootstrap --help | sed 's/\x1b\[[0-9;]*m//g' | awk 'ex==1 { gsub(/^[ $]*[ ]*/, ""); print; } /Example:/ { print "#!/usr/bin/env bash\n\nset -euo pipefail\n"; ex = 1; } END { print "END" }' | sed '{N;s/\(.*\)END/\nexpected=\1/;}' | bash
```

piotr-iohk:
```
cardano-address address bootstrap --help | sed 's/\x1b\[[0-9;]*m//g' | awk 'ex==1 { gsub(/^[ $]*[ ]*/, ""); print; } /Example:/ { print "#!/usr/bin/env bash\n\nset -euo pipefail\n"; ex = 1; } END { print "END" }' | sed '{N;s/\(.*\)END/\nexpected=\1/;}' | bash
```
:point_up_2: wow!

> Could we have this test all in Hspec though? It would fit in quite well with the existing tests.

I guess we could. One extra point however for having it outside is that, aside from testing user facing examples, it is at the same time an e2e test on already built binary, which makes it a nice, sort of, top of the test pyramid extra check... :sweat_smile: 

I think also those tests are pretty lightweight (adding ~1-2s to the CI time)... I don't think they would require any maintenance either, as long as new examples are added the same way as the existing ones.

rvl (approved):
I tried locally and noticed that the tests passed even when the `cardano-address` command was missing. So I fixed that and a couple other things, and added a little doc. Because I'm not very good with ruby, it does take much longer to set up and edit the tests than it normally should.

PR #128, opened 2021-04-21
Move jsbits from cabal.project to nix cabalProjectLocal

hamishmack:
This prevents the `cabal` from attempting to build the package inside the nix shell (instead the version provided by the nix shell will be used).

This prevents errors like:

```
[nix-shell:~/iohk/cardano-addresses]$ js-unknown-ghcjs-cabal run cardano-addresses:test:unit
Warning: don't know how to find change monitoring files for the installed
package databases for ghcjs
Warning: Requested index-state 2021-03-23T15:14:00Z is newer than
'hackage.haskell.org'! Falling back to older state (2021-03-23T15:08:40Z).
Resolving dependencies...
Build profile: -w ghcjs-8.10.4 -O1
In order, the following will be built (use -v for more details):
 - cardano-addresses-jsbits-3.3.0 (lib) (first run)
 - cardano-addresses-3.3.0 (lib) (configuration changed)
 - cardano-addresses-3.3.0 (test:unit) (configuration changed)
Configuring library for cardano-addresses-jsbits-3.3.0..
Preprocessing library for cardano-addresses-jsbits-3.3.0..
Building library for cardano-addresses-jsbits-3.3.0..
[1 of 1] Compiling Cardano.Address.Jsbits ( lib/Cardano/Address/Jsbits.hs, /home/hamish/iohk/cardano-addresses/dist-newstyle/build/js-ghcjs/ghcjs-8.10.4/cardano-addresses-jsbits-3.3.0/build/Cardano/Address/Jsbits.js_o )
jsbits/cardano-crypto.js: openBinaryFile: does not exist (No such file or directory)
cabal: Failed to build cardano-addresses-jsbits-3.3.0 (which is required by
test:unit from cardano-addresses-3.3.0).
```

The only drawback is that if we need to use `cabal` to build with ghc outside of a nix-shell it will need a modified `cabal.project` file or `cabal.project.local` file added.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -31,6 +31,7 @@ source-repository-package
 
 -- Relax overly strict bounds in hjsonschema and hjsonpointer
 allow-newer:
-    hjsonschema:aeson
-  , hjsonschema:hjsonpointer
-  , hjsonpointer:aeson
+    hjsonschema:*
+  , hjsonpointer:*
+
+test-show-details: direct
--- a/cabal.project
+++ b/cabal.project
@@ -7,7 +7,6 @@ with-compiler: ghc-8.6.5
 packages:
   command-line/cardano-addresses-cli.cabal
   core/cardano-addresses.cabal
-  jsbits/cardano-addresses-jsbits.cabal
 
 -- Needed for ghcjs
 -- See https://github.com/haskell-crypto/cryptonite/pull/345
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,4 @@
-index-state: 2021-03-23T15:14:00Z
+index-state: 2021-03-23T15:08:40Z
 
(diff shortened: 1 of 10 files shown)


rvl:
Oh right - I had a local copy of `jsbits/cardano-crypto.js`, so that's why it worked for me.

We can't assume that ghc users will have nix or be running nix-shell.
I would also like to limit CPP as much as possible.

What about a single module in the `core` package which has CPP? I pushed a change with this, but haven't tested much.

Also I noticed that the nix-shell requires a build of cardano-addresses. I suppose this is because of `build-tools: cardano-address` in one of the cabal files? It would be nice to fix that somehow because it makes the shell slow to start.

rvl (approved):

PR #134, opened 2021-05-31
Enable with/without chain code for private keys

paweljakubas:
Enabling `--with-chain-code` and `without-chain-code` in `key child` command


README.md
```diff
@@ -56,6 +56,25 @@ addr_xvk1grvg8qzmkmw2n0dm4pd0h3j4dv6yglyammyp733eyj629dc3z28v6wk22nfmru6xz0vl2s3
...
+
+```console
+$ cardano-address key child 1852H/1815H/0H/0/0 --with-chain-code < root.xsk
+addr_xsk1ppxa0t64fwwgclj6wa7u228d2ux06dkmx27zhc6eessgmj2849pn8k40xgjk5cd7jaw377gg558tlt5kgkfsnyr00mzqkxtfg5qn00wd8t99f5a37drp8k04ggj2rfkjx60y708m34n6udptx0hrdykllv9n5zcs
+
+$ cardano-address key child 1852H/1815H/0H/0/0 < root.xsk
+addr_xsk1ppxa0t64fwwgclj6wa7u228d2ux06dkmx27zhc6eessgmj2849pn8k40xgjk5cd7jaw377gg558tlt5kgkfsnyr00mzqkxtfg5qn00wd8t99f5a37drp8k04ggj2rfkjx60y708m34n6udptx0hrdykllv9n5zcs
+
+$ cardano-address key child 1852H/1815H/0H/0/0 --without-chain-code < root.xsk
+addr_xsk1ppxa0t64fwwgclj6wa7u228d2ux06dkmx27zhc6eessgmj2849pn8k40xgjk5cd7jaw377gg558tlt5kgkfsnyr00mzqkxtfg5qn00ga9r8zv
```
rvl, line 70: The bech32 HRP probably shouldn't have "x" if it's not an extended key.

paweljakubas: well I was also, at first glance, thinking it should be without `x`. Talked with Matthias about and the conclusion was that it should be always `xsk`.The private key is actually extended through hashing. It's 64 bytes long (so 96 with chain code), vs 32 normally.Unlike public keys, the extended secret keys are not compatible with normal Ed25519 secret keys at all.

paweljakubas: @KtorZ  correct?

KtorZ: The thing is, we have kind of _abused_ the "extended" notation for public keys. There's no such thing as an extended public key in the first place. There are public keys (32 bytes) and chain codes (32 bytes). Because of language abuse, we call "extended public keys" the combination of public key + chain code. 

Extended private keys however really are an extension of an actual Ed25519 private key. The key is extended through hashing, going from 32 bytes to 64 bytes. And, as for the public counterpart, it may or may not be associated with a chain-code. The chain code is only needed for key derivation, but not for signing.

Thus, we have to make a choice between:

- Calling _xsk an extended key, with or without chain code. 
- Calling _sk an extended key, without chain code. 

Both are wrong / misleading in a way, but I would argue that the latter is _more_ wrong. Ideally, we would need 3 prefixes:

- One for non extended secret keys (e.g. `*_sk`)
- One for extended secret keys without chain code (e.g. `*_xsk`)
- One for extended secret keys with chain code (e.g. `*_xccsk`) 

However, if we do something for extended secret keys w.r.t chain code, we have to be consistent with the public counterpart (e.g. `_*sk` & `*_ccsk`). There's no way to resolve this properly without a breaking change (which is arguably pretty low impact but may confuse a few people).

KtorZ: Note that we _could_ introduce this change in a backward-compatible way by letting the parser succeeds with the old prefixes, while using the new ones in outputs.

paluh:
@paweljakubas could you please tell me why this feature was not added to the tool?

PR #135, opened 2021-06-02
Initial version of cardano-addresses-api for ghcjs

hamishmack:
### Ticket

ADP-787

### Description

To try it out run:

```
nix-shell
js-unknown-ghcjs-cabal build cardano-addresses-api:unit
node dist-newstyle/build/js-ghcjs/ghcjs-8.10.4/cardano-addresses-api-3.4.0/t/unit/build/unit/unit.jsexe/all.js
```

That test initializes the api from a JS function that is called from Main.hs.  To build `.js` file that might be easier to use from a JS app run:

```
nix-build -A ghcjs-api -o api.js
```

This replaces the regular `runmain.js` with `api/jsbits/runmain.js`.  That exposes a single function you can call and in a continuation.

To initialize call the `runCardanoAddressesApi` with a continuation that like the one in `api/jsbits/test.js`.  You will be passed `api` and `cleanup`.

We might need @luite to advise on how best to refactor it so that the `runCardanoAddressesApi` can return (rather than running forever).

### Comments

[Hydra jobset](https://hydra.iohk.io/jobset/Cardano/cardano-addresses-pr-135#tabs-evaluations)


hamishmack:
> We might need @luite to advise on how best to refactor it so that the runCardanoAddressesApi can return (rather than running forever).

Actually I think it should be fine.  The call to `h$main(h$mainZCZCMainzimain);` does return before doing any work.  It does not wait for haskell land to terminate.  So `runCardanoAddressesApi` should work fine.  Just call it once and then save `api` value.  You might not want to bother with `cleanup` as it will also cause `process.exit` to be called on `node`.

api/lib/Cardano/Address/Api/InspectAddress.hs
```diff
@@ -0,0 +1,70 @@
...
+
+-- used for 'inspect'
+instance Error SomeException
+
+export :: Object -> JSM ()
+export api =
+    void $ api ^. jss "inspectAddress" (fun $ \ _ _ [rootPublicKeyJSVal, addressJSVal, onSuccess, onFail] -> do
+      mbRootPublicKeyText :: Maybe String <- fromJSValUnchecked rootPublicKeyJSVal
+      addressText :: Text <- fromJSValUnchecked addressJSVal
+      case mapM (xpubReader [CIP5.root_xvk]) mbRootPublicKeyText of
```
rvl, line 56: What are the options for making a wrapper "type" and parser for bech32-encoded bytes?

Say we have a function:
```
export function parseXPub(xpub: String): Promise<XPub> {
  // uses `xpubReader [CIP5.root_xvk]` to produce bytes
}
```

Would something like that work?

rvl (commented):
Awesome thanks @hamishmack!
I will try and find some time later to review fully so we can get the first version merged.

rvl:
@hamishmack, I couldn't resist adding some typescript wrappers and test cases. Try it with
```
cd api
nix-shell --run "npm install && npm run build && npm run test"
```
The tests don't work though, because I don't quite know how to do the ghc rts init process so that the API can be used as a CommonJS module. Any ideas on that?

jsapi/lib/Cardano/Address/JSAPI/InspectAddress.hs
```diff
@@ -0,0 +1,75 @@
...
+          case getBytes addressText of
+            Left e ->
+              void $ call onFail onFail [e]
+            Right bytes ->
+              case Shelley.eitherInspectAddress rootPublicKey (unsafeMkAddress bytes) of
+                Right json -> do
+                  jsVal <- toJSVal_aeson json
+                  void $ call onSuccess onSuccess [jsVal]
+                Left  e ->
+                  void $ call onFail onFail [displayException e]
```
rvl, line 74: Is it possible to convert this to a structured error type?
Perhaps we could just add an `instance ToJSON ErrInspectAddress`

rvl (approved):
I renamed some things and adjusted the `cardano-addresses.js` build a little.
This would be mergeable, except that the jest tests can't exit.

jsapi/lib/Cardano/Address/JSAPI/InspectAddress.hs
rvl: Actually don't worry about this for now - we can fix it later.

PR #139, opened 2021-06-24
Update badges for Linux, Windows and Typescript builds

piotr-iohk:
ADP-787

-> https://github.com/input-output-hk/cardano-addresses/tree/piotr/typescript-badge

Perhaps that would be nice to have too...


rvl (approved):
Good idea, thanks.
I also added a badge for NPM, and updated docs and package version ready for release there.

PR #140, opened 2021-06-25
Add NFData instances to Cardano.Mnemonic types

rvl:
## Issue

ADP-909

## Description

Adds NFData instances so that we can more easily derive NFData for types which include these types.


jonathanknowles (approved):
Looks good to me!

PR #141, opened 2021-06-25
remove deprecated ALLOW_TAG_PREFIX from release workflow

piotr-iohk:
It seems it is deprecated (https://github.com/input-output-hk/cardano-addresses/runs/2911314075?check_suite_focus=true).
I'm not sure if we need it at all so removing all together.

If needed, then can be replaced by something like `TAG_PREFIX_REGEX: [a-z-]*` as per https://github.com/anton-yurchenko/git-release#configuration.


rvl (approved):
Thanks. Since the tags don't have any prefix in this repo, it should work just fine.

PR #144, opened 2021-06-30
Use latest ghcjs

hamishmack:
This includes https://github.com/ghcjs/ghcjs/commit/2da23efdeb02b0570fd0693ec1610287f785f042 and https://github.com/ghcjs/ghcjs/commit/e4cd4232a31f6371c761acd93853702f4c7ca74c to fix the duplicates in `rts.js`.

### DIFF
--- a/nix/cardano-addresses-js.nix
+++ b/nix/cardano-addresses-js.nix
@@ -28,10 +28,6 @@ stdenv.mkDerivation rec {
       diff -Nur "$js".bak "$js" || true
     }
 
-    sed_inplace rts.js \
-      -e '0,/function h\$ap_1_0(/s//function extra_h\$ap_1_0(/' \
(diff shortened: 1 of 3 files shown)


rvl (approved):
Sweet, thanks!

rvl:
It looks like the ghcjs build is failing at the `lens` dependency, @hamishmack. I wonder why it would fail like that? If we set index-state, most versions should stay this same.

luite:
I don't think the `lens` failure is related to this pull request. Rerunning the job might well fix it. Unfortunately there's still a bug in ghcjs that causes intermittent typechecker failures like this.

I hope we can fix that soon, since this is quite annoying, especially in situations with CI.

hamishmack:
@rvl restarting the jobs on hydra did seem to fix this one.

rvl:
OK thanks, yes I also hope we can fix this intermittent typechecker failure issue in ghcjs.

PR #155, opened 2021-10-04
Flakify nix build. [WIP]

jbgi:


disassembler (approved):

sevanspowell:
Thanks @jbgi! I haven't yet had a reason to look into flakes, so my approval doesn't mean a whole lot, but if it builds, great!

My only concern, and possibly it's a blocker, is the upgrade of ghc. This issue has been mentioned a few times in different contexts, but we've never found an answer. Local Nix builds will work, but Buildkite will fail with: `version 'GLIBC_2.33' not found`. See:
- https://input-output-rnd.slack.com/archives/CAP8NM7N0/p1631723169051000
- https://input-output-rnd.slack.com/archives/CAP8NM7N0/p1629467921080500
- and my current PR https://github.com/input-output-hk/cardano-wallet/pull/2947, which fails for the same reason on Buildkite.

So I think this branch is only passing CI because it doesn't have a Buildkite job. Do you think we can overcome this issue with Buildkite?

sevanspowell (approved):

PR #157, opened 2021-12-13
Add address type to inspect

paweljakubas:
[DDW-742](https://input-output.atlassian.net/browse/DDW-742)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ $ echo addr_test1vp2fg770ddmqxxduasjsas39l5wwvwa04nj8ud95fde7f7guscp6v | cardano
     "address_style": "Shelley",
     "spending_key_hash": "54947bcf6b760319bcec250ec225fd1ce63baface47e34b44b73e4f9",
     "network_tag": 0,
-    "address_type": "0110"
+    "address_type": 6
 }
 
 $ echo addr_test1qp2fg770ddmqxxduasjsas39l5wwvwa04nj8ud95fde7f70k6tew7wrnx0s4465nx05ajz890g44z0kx6a3gsnms4c4qq8ve0n | cardano-address address inspect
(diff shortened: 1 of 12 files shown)


core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -522,65 +524,75 @@ parseAddressInfoShelley AddressParts{..} = case addrType of
             { infoStakeReference = Just ByValue
             , infoSpendingKeyHash = Just addrHash1
             , infoStakeKeyHash = Just addrHash2
+            , infoAddressType = "0000"
```
KtorZ, line 527: Wouldn't it make more sense to show the base-10 value of the type? i.e. here type `0`

paweljakubas: yes, but then icarus and byron address will not be differentiable only looking at this field..both assume "1000"

KtorZ: But how would showing the bit sequence make it any better for Icarus and Byron :no_mouth: ?

paweljakubas: agree, fix applied https://github.com/input-output-hk/cardano-addresses/pull/157/commits/ae2a6ff87846d615f801ac4e7430872f75563d97

core/lib/Cardano/Address/Style/Shelley.hs
```diff
@@ -522,65 +522,75 @@ parseAddressInfoShelley AddressParts{..} = case addrType of
             { infoStakeReference = Just ByValue
             , infoSpendingKeyHash = Just addrHash1
             , infoStakeKeyHash = Just addrHash2
+            , infoAddressType = 0
```
rvl, line 525: Rather than hard-coding numbers, can we just get `infoAddressType` using a bitwise "and" and right shift?

README.md
```diff
@@ -99,6 +99,50 @@ stake_test1urmd9uh08pen8c26a2fn86weprjh52638mrdwc5gfac2u2s25zpat%
...
+    "spending_key_hash_bech32": "addr_vkh12j28hnmtwcp3n08vy58vyf0arnnrhtavu3lrfdztw0j0jng3d6v",
+    "address_style": "Shelley",
+    "spending_key_hash": "54947bcf6b760319bcec250ec225fd1ce63baface47e34b44b73e4f9",
+    "network_tag": 0,
+    "address_type": 0
+}
+```
+
+Details about possible address types are following.
+| address_type | binary prefix  |   Meaning                                                |
```
rvl, line 130: Address type and binary prefix columns are the same number, just decimal/binary representations, right?

Can we add a reference to the shelley CDDL, if none present yet?

rvl (approved):
Looks good, thanks @paweljakubas. A couple review comments below ...

rvl:
PR #158 will fix the CI failures - unrelated to this PR.

core/lib/Cardano/Address/Style/Shelley.hs
paweljakubas: good idea -> done

PR #166, opened 2022-01-20
Introduce wallet id in cli and add hex-encoded option to hashing of keys 

paweljakubas:
ADP-1359

Two things were done.
1. It was added "key walletid" that works on EXTENDED root and account keys, both private and public, to get wallet id.
A number of tests were added.

2. Optional flag "--hex" was added to "key hash" command. If used, then instead of bech32-encoded output user gets hex-encoded one.

### DIFF
--- a/command-line/lib/Command/Key/WalletId.hs
+++ b/command-line/lib/Command/Key/WalletId.hs
@@ -1,0 +1,80 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+module Command.Key.WalletId
+    ( Cmd (..)
+    , mod
+    , run
+    ) where
+
+import Prelude hiding
+    ( mod )
+
+import Cardano.Address.Derivation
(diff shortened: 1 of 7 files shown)


command-line/lib/Command/Key/WalletId.hs
```diff
@@ -0,0 +1,79 @@
...
+
+data Cmd = WalletId
+    deriving (Show)
+
+mod :: (Cmd -> parent) -> Mod CommandFields parent
+mod liftCmd = command "walletid" $
+    info (helper <*> fmap liftCmd parser) $ mempty
+        <> progDesc "Get the wallet id from a given key"
+        <> footerDoc (Just $ string $ mconcat
+            [ "The key is read from stdin."
```
rvl, line 42: Could we add some info for users explaining what a wallet id is?

Also mention that users can supply either a vk or sk - if a sk is supplied then the public key part will be used.

command-line/lib/Command/Key/WalletId.hs
```diff
@@ -0,0 +1,79 @@
...
+            ])
+  where
+    parser = pure WalletId
+
+run :: Cmd -> IO ()
+run WalletId = do
+    (hrp, bytes) <- hGetBech32 stdin allowedPrefixes
```
rvl, line 49: Perhaps in a future version we could accept hex-encoded vks?
I know this would not be able to validate the type of key, but it might be helpful for users to have a back door.

rvl (approved):
Nice

PR #178, opened 2022-04-02
Changes to compile with both aeson 1 & 2

newhoggy:


rvl:
Thanks @newhoggy - a PR description would be helpful.

piotr-iohk:
Adressed in #182.

PR #179, opened 2022-04-02
Aeson compatibility

newhoggy:


.github/workflows/haskell.yml
```diff
@@ -0,0 +1,47 @@
+name: Binaries
```
rvl, line 1: Could this change please be moved to #178?

core/cardano-addresses.cabal
```diff
@@ -53,8 +53,8 @@ library
       NoImplicitPrelude
   ghc-options: -Wall -Wcompat -fwarn-redundant-constraints
   build-depends:
-      aeson
-    , base >=4.7 && <5
+      base >=4.7 && <5
+    , aeson ^>= 1.5.6.0 || ^>= 2.0.3.0
```
rvl, line 57: For this library, I don't think we have any particular need to support both aeson versions.

rvl (changes requested):
Thanks @newhoggy

core/cardano-addresses.cabal
newhoggy: We have quite a lot of code that relies on `aeson-1`.  I think it will be less disruptive to downstream developers if we support both versions on a very temporary basis.

newhoggy:
I have moved this change to #178 as requested.  Closing.

PR #182, opened 2022-04-07
Aeson 1 & 2 compatibility

newhoggy:
Make it possible to compile with both versions of `aeson` so downstream projects can be on whatever they need.

This change is meant to be temporary.  A full upgrade to `aeson-2` would occur when all downstream projects have completed their migration.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -1,0 +1,53 @@
+name: Binaries
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: ["8.10.7"]
+        os: [ubuntu-latest]
+        aeson: ["1.5.6.0", "2.0.3.0"]
+        exclude:
+          - ghc: "9.2.2"
+            aeson: "1.5.6.0"
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: haskell/actions/setup@v1
+        id: setup-haskell
+        with:
+          ghc-version: ${{ matrix.ghc }}
+          cabal-version: 3.6.2.0
+
+      - name: Configure project
+        run: |
+          # We can't update the index state yet because nix flake is locked to an earlier one.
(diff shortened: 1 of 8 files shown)


cabal.project
```diff
@@ -31,10 +31,21 @@ source-repository-package
...
     type: git
-    location: https://github.com/ghcjs/jsaddle
-    tag: 9cb3534b6b2dc948c6fa88fce281e59e904b43e4
-    --sha256: 1drgrgjnxamq8vh6mrqg62xcj7vb9makpy89bls485d6ckigv3ha
-    subdir: jsaddle
+    location: https://github.com/input-output-hk/hjsonpointer
+    tag: bb99294424e0c5b3c2942c743b545e4b01c12ce8
+    --sha256: 11z5s4xmm6cxy6sdcf9ly4lr0qh3c811hpm0bmlp4c3yq8v3m9rk
+
+source-repository-package
```
michaelpj, line 38: Please add comments explaining what these are for and when they can be removed.

.github/workflows/haskell.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        ghc: ["8.10.7"]
+        os: [ubuntu-latest]
+        aeson: ["1.5.6.0", "2.0.3.0"]
```
michaelpj, line 22: Add a comment explaining what this is for, namely testing that we do actually have aeson 1/2 compatibility.

michaelpj (approved):
Code looks fine, but you need to explain things!

hamishmack (approved):
Looks good.  I tried updating `haskellNix` and `index-state` to check building with aeson 2 and noticed that the new nix filtering in Haskell.nix seems to break the building of the ghcjs lib (at least with Nix 2.5 on linux).  I'll look into that.

.github/workflows/haskell.yml
newhoggy: Explanations added.

cabal.project
newhoggy: Explanations added.

rvl:
Out of the 5 PRs #178 #179 #180 #182 #183 related to aeson 2, which one are we supposed to be reviewing and merging?

sevanspowell:
Rebased on master.

PR #185, opened 2022-05-16
Walletid support for shared wallets

paweljakubas:
ADP-1621

adding support for walletid calculation for multisig wallets

### DIFF
--- a/command-line/lib/Command/Key/WalletId.hs
+++ b/command-line/lib/Command/Key/WalletId.hs
@@ -35,6 +35,8 @@ import System.IO.Extra
 
 import qualified Cardano.Codec.Bech32.Prefixes as CIP5
 import qualified Data.ByteString as BS
+import qualified Data.Text as T
+import qualified Data.Text.Encoding as T
 
 data Cmd = WalletId
     { spending :: Script Cosigner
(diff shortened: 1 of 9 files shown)


command-line/lib/Command/Key/WalletId.hs
```diff
@@ -15,62 +16,98 @@ import Prelude hiding
...
         , CIP5.root_xvk
-        , CIP5.root_shared_xsk
-        , CIP5.root_shared_xvk
         , CIP5.acct_xvk
         , CIP5.acct_xsk
         , CIP5.acct_shared_xvk
         , CIP5.acct_shared_xsk
         ]
 
+    scriptTemplateToBs = T.encodeUtf8 . T.pack . show
```
HeinrichApfelmus, line 107: Hm, using `show` here will probably make it very difficult for non-Haskell code to generate matching wallet IDs 🤔. Then again, wallet IDs are an internal affair, so we don't have to cater to that.

Still, I was wondering whether we should use `serializeScript` to get a canonical serialization before hashing, but this function expects a `Script KeyHash`, not `Script Cosigner`.

Proposal: How about renaming `scriptTemplateToBs` to
```hs
serializeScriptTemplate :: Script Cosigner -> ByteString
```
and exporting it on the top level from `Cardano.Address.Script`? In this way, the function that serializes a script template is at least "fully specified and exported" (and we can later reuse it in cardano-wallet).

HeinrichApfelmus (approved):
LGTM! 😊 But I would propose to export `serializeScriptTemplate` to make the serialization of script templates more "official" (as the serialization is important to get the right hash hash).

command-line/lib/Command/Key/WalletId.hs
paweljakubas: good idea.

1. now we will calculate byte payload by utf8 encoding of strings that are used in json/command line. so
```
cosigner#0
all [cosigner#1, cossigner#0, active_from 1000]
```
rather than show instance. It will be easier to reconstruct it for any user doing this outside cardano-addresses

2. I have exposed `sharedWalletId` to be used in cardano-wallet to promote better reuse of cardano-addresses and avoid unneeded redundancy

PR #189, opened 2022-06-07
Upgrade to aeson 2

Anviking:
Needed from https://github.com/input-output-hk/cardano-wallet/pull/3280

### Comments

I now see there's the existing PR #182, but this is more minimal.

... However CI here doesn't seem to pass...


piotr-iohk:
Adressed in #182.

PR #191, opened 2022-07-07
Extend payment share key derivation

paweljakubas:
ADP-1623

Allowing to derive payment address, both private and public, for both UTxOInternal and UTxOExternal. It is needed in multisig work in cardano-wallet


HeinrichApfelmus (approved):
Looks good to me! 😊

PR #194, opened 2022-08-09
Bump hw-aeson to v0.1.6.0

sevanspowell:
- cardano-node 1.35.3 is using hw-aeson v0.1.6.0.
- v0.1.6.0 of hw-aeson changes the API such that it no longer exposes "toList",
  but only when compiled with versions of aeson > 2.0.0.0.
  - This can be a bit confusing, as cardano-addresses uses aeson < 2.0.0.0 by
  default.
- Change how cardano-addresses uses the hw-aeson API to workaround this breakage
  in the API.
  - Simply re-write "toList" in terms of "foldlWithKey'".
  - Both `toList` and `foldlWithKey'` have O(n) complexity.

### DIFF
--- a/core/lib/Cardano/Address/Script.hs
+++ b/core/lib/Cardano/Address/Script.hs
@@ -801,9 +801,11 @@ instance FromJSON ScriptTemplate where
         ScriptTemplate <$> (Map.fromList <$> cosigners') <*> template'
       where
         parseCosignerPairs = withObject "Cosigner pairs" $ \o ->
-            case JM.toList o of
+            case toList o of
                 [] -> fail "Cosigners object array should not be empty"
                 cs -> for (reverse cs) $ \(numTxt, str) -> do
                     cosigner' <- parseJSON @Cosigner (String (J.keyToText numTxt))
                     xpub <- parseXPub str
                     pure (cosigner', xpub)
+
+        toList = JM.foldlWithKey' (\acc k v -> (k,v):acc) []
--- a/cabal.project
+++ b/cabal.project
@@ -31,7 +31,7 @@ source-repository-package
 
 -- TODO This was patched to work with aeson 1 & 2 with the help of hw-aeson.
 -- Once downstream projects are all upgraded to work with aeson-2, they can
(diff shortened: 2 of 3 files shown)


core/lib/Cardano/Address/Script.hs
```diff
@@ -801,9 +801,11 @@ instance FromJSON ScriptTemplate where
...
         parseCosignerPairs = withObject "Cosigner pairs" $ \o ->
-            case JM.toList o of
+            case toList o of
                 [] -> fail "Cosigners object array should not be empty"
                 cs -> for (reverse cs) $ \(numTxt, str) -> do
                     cosigner' <- parseJSON @Cosigner (String (J.keyToText numTxt))
                     xpub <- parseXPub str
                     pure (cosigner', xpub)
+
+        toList = JM.foldlWithKey' (\acc k v -> (k,v):acc) []
```
paweljakubas, line 811: I think we could remove the imports
```
import qualified HaskellWorks.Data.Aeson.Compat as J
import qualified HaskellWorks.Data.Aeson.Compat.Map as JM
```
and use https://hackage.haskell.org/package/aeson-2.1.0.0/docs/Data-Aeson-KeyMap.html
(KeyMap is Foldable)

paweljakubas: possibly we don't need hw-aeson at all, and use functionality exposed from `Data.Aeson.KeyMap`

sevanspowell: `Data.Aeson.Keymap` isn't exported in aeson versions < 2.0.0.0, which is why I think we're using `hw-aeson`, as a compatibility shim.

sevanspowell: And not all our users have migrated to Aeson 2 yet unfortunately.

paweljakubas: aha, I got it. Thanks for the explanation!

paweljakubas (approved):

sevanspowell:
Closed in favour of https://github.com/input-output-hk/cardano-addresses/pull/195

PR #202, opened 2022-08-19
Expose from recovery passphrase

paweljakubas:
https://input-output.atlassian.net/browse/ADP-2143

Adding second factor mnemonic support to root private key derivation.
Mnemonics having 9-word and 12 word are supported. Alternatively one can use base16/base64/utf8 encoded equivalent of them. Or just the array of byte octets.
Passing with "" is equivalent to not using it at all.

Docs updated.


command-line/lib/Options/Applicative/Style.hs
```diff
@@ -89,5 +123,35 @@ styleArg = argument (eitherReader reader) $ mempty
...
+            Left _ -> getBase16 <|> getBase58
+  where
+    getBase16 = case fromBase16 (T.encodeUtf8 $ T.pack str) of
+        Right hex -> pure $ FromEncoded hex
+        Left _ -> Left "not hex-encoded bytes"
+    getBase58 = case fromBase58 (T.encodeUtf8 $ T.pack str) of
+        Right bs -> pure $ FromEncoded bs
+        Left _ -> Left "not base58-encoded bytes"
+    noNewline :: Char -> Bool
+    noNewline = (`notElem` ['\n', '\r'])
```
KtorZ, line 146: I would suggest to not enforce anything particular on the 2nd-factor. Just a plain UTF-8 bytestring. There's no reason to force a particular ascii-friendly encoding nor to use mnemonic for that. I know that's what cardano-wallet's API does but that's a very debatable design decision -- especially in the context of the command-line.

README.md
```diff
@@ -45,11 +45,48 @@ exercise club noble adult miracle awkward problem olympic puppy private goddess
...
 </details>
 
+<details>
+  <summary>How to generate a root private key with passphrase (<strong>root.xsk</strong>)</summary>
+
+```console
+$ cardano-address recovery-phrase generate --size 9 > sndfactor.prv
+swing payment diagram happy chimney mammal flip become lyrics
+
+$ cardano-address key from-recovery-phrase Shelley --passphrase "$(cat sndfactor.prv)" < phrase.prv
```
KtorZ, line 65: You generally want to avoid passing sensitive stuff like passphrases as command-line options, as these remains in the shell history afterwards. I would suggest to offer two alternatives to users:

- Provide a passphrase via a file; 
- Or else, if no file is provided (but only `--passphrase`), enter the passphrase interactively.

paweljakubas:
@KtorZ thanks for looking at the PR. I will adopt the changes you suggested right after vacations (after 5th Sep) that I am just starting. Cheers!

command-line/lib/Options/Applicative/Style.hs
paweljakubas: I supported plain utf8 text, bytestring hex encoded, bytestring base64 encoded and octet array

README.md
paweljakubas: see the added section in Readme.md to see all supported options

piotr-iohk: So currently we have several options:
```
cardano-address key from-recovery-phrase Shared --passphrase from-mnemonic
cardano-address key from-recovery-phrase Shared --passphrase from-hex
cardano-address key from-recovery-phrase Shared --passphrase from-base64
cardano-address key from-recovery-phrase Shared --passphrase from-utf8
cardano-address key from-recovery-phrase Shared --passphrase from-octets
```
Each of those options triggers interactive mode accordingly, e.g.:
```
Please enter a [9, 12, 15, 18, 21, 24] word mnemonic:
skill consider antenna lunar arch coach coast symbol rent
Please enter passphrase in the form of octet array:
[220,56]
```
As far as I see there are two issues:

1. The interactive mode does not allow making any mistake... when user types wrong thing he's not able to correct it as hitting Backspace, "arrow" or del adds additional characters:
```
Please enter a [9, 12, 15, 18, 21, 24] word mnemonic:
ere ere ^[[D^[[D^?^?^[[3~^[[3~^[[C^[[C
```

2. Currently there's no non-interactive mode that would allow user to set `--passphrase` without going into interactive mode. I believe that non-interactive mode would be quite useful if one would like to use --passphrase feature in some automation workflow:

I think that what @KtorZ proposes makes sense:
>Provide a passphrase via a file;

which would enable non-interactive mode, e.g.:
```
cardano-address recovery-phrase generate \
| cardano-address key from-recovery-phrase Shared --passphrase /path/to/phrase.prv
```

> Or else, if no file is provided (but only --passphrase), enter the passphrase interactively (with confirmation)

Which should be good calling it like this:
```
cardano-address recovery-phrase generate \
| cardano-address key from-recovery-phrase Shared --passphrase
Please enter additional passphrase in the form of utf-8:
```
or like this:
```
cardano-address key from-recovery-phrase Shared --passphrase
Please enter a [9, 12, 15, 18, 21, 24] word mnemonic:
...
Please enter additional passphrase in the form of utf-8:
...
```

I also think that from practical sense having options: from-mnemonic, from-hex, from-base64, from-octet is not particularly useful. It seems that plain utf-8 should be enough for additional `--passphrase`.

paweljakubas: well, `from-mnemonic` is definitely useful as we want seamless compatibility with cardano-wallet which uses mnemonic as second factor in api. The rest are optional. One needs to remember that not always mnemonic passphrase maps to utf8 passphrase (but aways to base16 or base64 and of course octet). Nevertheless, I see the point of adding possibility to read from file :+1:

paweljakubas: @piotr-iohk So now we have the following:
1. we have backspace support in interactive mode
2. we have explicit, sensitive and silent mode 
3. we have support for all from-format modes in file mode
4. docs updated
5. added comprehensive testing for from-file

paweljakubas:
we have the same what was reported here https://github.com/mstksg/setup-stack/issues/16

piotr-iohk (approved):
LGTM, CI issues seem resolved.

PR #204, opened 2022-09-29
Attempt to fix build

piotr-iohk:


piotr-iohk:
Closing, resolved directly in `paweljakubas/adp-2143/expose-from-recovery-passphrase`.

PR #206, opened 2022-10-23
Use environment files in Github Actions

newhoggy:
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


paweljakubas:
applied already in https://github.com/input-output-hk/cardano-addresses/pull/215

PR #212, opened 2022-12-08
Add Unknown key role and adjust code needed

paweljakubas:
this is part of ADP-2379. `KeyRole` was extended and `Unknown` data constructor was added. And the needed code adjusted. The change is motivated by the fact that there are situations where for a given wallet we cannot state whether a given key hash is acting as Policy/Payment/Delegation, it is external for this wallet and unknown. The keys can be represented as bytes and only upon constructing value of type KeyHash we are assuming the role. In order to handle the situation where the key hash is unknown we introduced the Unknown role. In that case we will assume hex-encoded representation - in contrast to bech32 where we the role is known.

https://input-output.atlassian.net/browse/ADP-2312

### DIFF
--- a/core/lib/Cardano/Address/Script.hs
+++ b/core/lib/Cardano/Address/Script.hs
@@ -221,7 +221,7 @@ scriptHashFromBytes bytes
     | BS.length bytes /= credentialHashSize = Nothing
     | otherwise = Just $ ScriptHash bytes
 
-data KeyRole = Payment | Delegation | Policy
+data KeyRole = Payment | Delegation | Policy | Unknown
     deriving (Generic, Show, Ord, Eq)
 instance NFData KeyRole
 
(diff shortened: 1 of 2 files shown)


paweljakubas:
Two things are needed:
1. Remove cicero
2. Fix Continuous Integration (Windows)
```
network                    >                  from cbits\asyncAccept.c:6:0: error: 
network                    > C:\Users\runneradmin\AppData\Local\Programs\stack\x86_64-windows\ghc-8.10.7\lib/include/stg/Types.h:26:9: warning: #warning "Mismatch between __USE_MINGW_ANSI_STDIO definitions. If using Rts.h make sure it is the first header included." [-Wcpp]
network                    >    26 | #       warning "Mismatch between __USE_MINGW_ANSI_STDIO definitions. \
network                    >       |         ^~~~~~~
network                    > In file included from C:\Users\runneradmin\AppData\Local\Programs\stack\x86_64-windows\ghc-8.10.7\lib/include/HsFFI.h:20,
```

Anviking (approved):
👌 lgtm

PR #215, opened 2022-12-12
Fix Windows CI

newhoggy:
Adapted from https://github.com/input-output-hk/ouroboros-network/pull/4159

ticket with description https://input-output.atlassian.net/browse/ADP-2499


paweljakubas (approved):
thanks @newhoggy ! We are now green, non-passing workflows working, repo reset with reasonable Required

PR #217, opened 2022-12-17
Fix caching in CI

newhoggy:


paweljakubas (approved):
thanks @newhoggy  . on my side I also updated CI badges as they were staled. Now we have the repo in very good shape :tada:

PR #230, opened 2023-09-08
Boost byron icarus docs

paweljakubas:
recheck and revisit code usage docs for byron and icarus

### DIFF
--- a/core/lib/Cardano/Address/Style/Byron.hs
+++ b/core/lib/Cardano/Address/Style/Byron.hs
@@ -205,18 +205,24 @@ type family DerivationPath (depth :: Depth) :: Type where
 -- > :set -XOverloadedStrings
 -- > :set -XTypeApplications
 -- > :set -XDataKinds
+-- > :set -XFlexibleContexts
 -- > import Cardano.Mnemonic ( mkSomeMnemonic )
+-- > import Cardano.Address ( base58 )
+-- > import Cardano.Address.Derivation ( toXPub )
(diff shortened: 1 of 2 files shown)


core/lib/Cardano/Address/Style/Byron.hs
```diff
@@ -205,18 +212,25 @@ type family DerivationPath (depth :: Depth) :: Type where
...
 -- === Deriving child keys
 --
--- > let Just accIx = fromWord32 0x80000000
--- > let acctK = deriveAccountPrivateKey rootK accIx
+-- > let Just accIx = indexFromWord32 0x80000000
+-- > let acctK = Byron.deriveAccountPrivateKey rootK accIx
 -- >
--- > let Just addIx = fromWord32 0x80000014
--- > let addrK = deriveAddressPrivateKey acctK addIx
+-- > let Just addIx = indexFromWord32 0x80000014
```
Unisay, line 229: I wonder if there is a reason to use exactly this number (or any hardened index will do) ?
it would be nice to clarify.  

```suggestion
-- > let firstHardenedIndex = 0x80000000
-- > let Just addIx = indexFromWord32 (firstHardenedIndex + 14)
```

paweljakubas: added requested information in examples with exact ranges possible for addIx and acctIx in both byron and icarus cases

Unisay (approved):

PR #231, opened 2023-09-12
Make it build with ghc-9.6

erikd:


angerman:
- All the macOS builds are broken due to changes in GH Runners.  My suggestion is to use the iohk/devx actions instead. They support those runners just fine.

- The Coding Standards check fails, but that can be fixed. Looks like mostly stylish issues.

- TypeScript NPM Package failes due to: a missing /home/runner/work/cardano-addresses/cardano-addresses/default.nix, which I believe we want to adjust.

aleeusgr:
I notice that some of inputs are set to track 21.05 channel of nixpkgs, while others track 23.05 
Should this be fixed?
https://github.com/IntersectMBO/cardano-addresses/blob/78be3215a1fd93567b481d91bdee8abd572952e5/nix/flake-compat.nix#L8-L12

angerman (approved):

.github/workflows/haskell.yml
```diff
@@ -12,13 +12,9 @@ jobs:
...
       fail-fast: false
       matrix:
-        ghc: ["8.10.7"]
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        aeson: ["2.0.3.0"]
-        include:
-          - ghc: "8.10.7"
-            os: ubuntu-latest
-            aeson: "1.5.6.0"
+        ghc: ["8.10.7", "9.2.8", "9.6.5"]
```
angerman, line 15: We may want to consider dropping 9.2.8?

.github/workflows/haskell.yml
```diff
@@ -172,9 +168,13 @@ jobs:
...
 
     - name: Run unit tests
+      shell: bash
       run: |
         cabal test cardano-addresses
-        cabal test cardano-addresses-cli
+        if [ "${{ runner.os }}" != "Windows" ] && [ "${{ matrix.ghc }}" !=  "9.6.2" ]; then
+          # This test is currently broken on windows (unhandled PEi386 relocation type 14) so disable it.
+          cabal test cardano-addresses-cli
+        fi
```
angerman, line 177: I _think_ this is fixed in recent haskell.nix versions.

.github/workflows/haskell.yml
Crypto2099: Should we move this out to an issue after getting this merged or is it something we should address prior to proceeding with next steps of getting the rest of the build processes back to a working state @disassembler?

.github/workflows/haskell.yml
Crypto2099: Can also be moved to a separate issue after this is merged and we can move the repository forward?

command-line/lib/Command/Key/WalletId.hs
```diff
@@ -57,8 +59,8 @@ mod :: (Cmd -> parent) -> Mod CommandFields parent
 mod liftCmd = command "walletid" $
     info (helper <*> fmap liftCmd parser) $ mempty
         <> progDesc "Shows the cardano-wallet wallet ID for a given key"
-        <> footerDoc (Just $ string $ mconcat
-            [ "A wallet ID is a 40-digit hexadecimal string derived "
+        <> footerDoc (Just $ pretty $ mconcat
+            [ "A wallet ID is a 40-digit hexadecimal pretty derived " :: Text
```
Crypto2099, line 63:
```suggestion
            [ "A wallet ID is a 40-digit hexadecimal string derived " :: Text
```

Find and replace error by the looks of it :)

Crypto2099 (approved):
Looks like we are good to proceed with this as a first step

.github/workflows/haskell.yml
hamishmack: This is not doing anything since `"9.6.2" != "9.6.5"`

hamishmack: I've changed this back.

hamishmack: It turns out it was still filtering out the test as the `&&` should have been an `||`.  So it was disabling all the test for windows GHC versions (including `"9.6.5"`).  I have put it back with:
* `||` instead of `&&`
* `9.6.5` instead of `9.6.2`
* A comment as I think the failure is related to encoding of emoji's in the output.

PR #234, opened 2023-10-12
Use with utf8 and take care of hpack problems

paweljakubas:


jonathanknowles:
LGTM!

jonathanknowles (approved):

paweljakubas:
hi @dermetfan I have TypeScript NMP Package fail on CI. And it says

angerman:
@paweljakubas i'll have a look tomorrow.

tgunnoe:
> Do you know how to change this parameter?

https://github.com/IntersectMBO/cardano-addresses/blob/master/.github/workflows/typescript.yml#L16

tgunnoe:
ok, maybe the size of `/tmp` on the machine is running out of space. it would be good to evaluate if so and why that's happening.

paweljakubas:
@tgunnoe do you know HOW and WHERE to check /tmp?  Changing limit 3x seems not to work ... I will try to remove env in workflow

tgunnoe:
in the step https://github.com/IntersectMBO/cardano-addresses/pull/234/files#diff-49782368fc4afbd0b5be05e65cc0a4677d70c2078d3206dbe6b347b19d41630cR33
I would change it before the typedoc task is ran:
```
-      run: 'nix-shell --run "npm run typedoc"'
+      run: |
+        df -h /tmp
+        nix-shell --run "npm run typedoc"
```

paweljakubas:
Local inspection:
```
1M	./bin
1M	./demo
1M	./exe
1M	./glue
1M	./lib
1M	./src
1M	./test
101M	./dist
341M	./node_modules
442M	.
442M	total
```

tgunnoe:
Ah! I think i know the issue now.  its trying to compile haskell.nix's stuff from source, which is waaay over 80gb.  I think we need to update the nixpkgs version that haskell.nix follows. let me get back to you.

paweljakubas:
hi @tgunnoe @angerman any progress here? This seems to be blocking issue to do anything within this repo. We have security alert for example and the same issue -> https://github.com/IntersectMBO/cardano-addresses/pull/235
Thanks!

tgunnoe:
@paweljakubas Hey, sorry for the delay.
I tried it out on my own fork which doesnt use the nix shell provided by jsapi, because I thought it was only needing simple node scripts to be ran. It doesn't. I needs the api output of the stack to run the tests
https://github.com/tgunnoe/cardano-addresses/actions/runs/6669489321/job/18127376171#step:6:70
So I've asked @angerman if @hamishmack could help out with why on this PR after removing hpack you're needing to build the whole stack and not get cache hits.

paweljakubas:
hi @tgunnoe Can you kindly say how to do it? Thanks

paweljakubas:
thanks @angerman

PR #252, opened 2024-06-10
fix jsapi build

aleeusgr:
NPM Typescript action fails with error 
> Package key for wired-in dependency `ghcjs-th' could not be found. 

to reproduce:
1. ` cd jsapi`
2. `nix-shell --run "npm install"`

Potential reasons for this error:

1. Missing or outdated package cache: The package cache might not have the required package or its dependencies. 
3. Incorrect or outdated Nix configuration: Your Nix configuration might be pointing to an outdated or incorrect package set. 
4. Dependency conflict: There might be a conflict between different package versions or dependencies. 
5. GHCJS version issue: The error mentions ghcjs-8.10.7-BwWb2KVba9L98LVnLzrqIb, which might indicate a version issue with GHCJS. Ensure you're using a compatible version of GHCJS with your Nix setup.

ghcjs-th is a library in ghcjs package, it is available [in 8.10](https://github.com/ghcjs/ghcjs/tree/ghc-8.10/lib/ghcjs-th); 

try the following:

- [x]  update your Nix environment and ensure that all dependencies are up-to-date.
  - bump nixpkgs, 
  - flake update, 
  - bump flake-compat

- [x] Try a clean build: collect garbage, then rerun build.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,6 +16,23 @@
         "type": "github"
       }
     },
+    "blst": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1691598027,
+        "narHash": "sha256-oqljy+ZXJAXEB/fJtmB8rlAr4UXM+Z2OkDa20gpILNA=",
+        "owner": "supranational",
+        "repo": "blst",
+        "rev": "3dd0f804b1819e5d03fb22ca2e6fac105932043a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "supranational",
+        "ref": "v0.3.11",
+        "repo": "blst",
+        "type": "github"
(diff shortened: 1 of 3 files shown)


disassembler:
We need to get working CI again before merging this (and it might not be necessary). Main work is happening in #231 which needs a lot of upstream fixes that are currently being addressed. Any contributions that fix failing jobs in the PR would be very welcome if you have time.

aleeusgr:
@disassembler exploring why CI is failing is the goal of this PR. I was talking to Adam Dean about upgrading my nix-fu and they suggested for me to join this effort. Thanks for pointing towards where the action is.

PR #254, opened 2024-06-12
Enable JS GHA

hamishmack:


angerman:
Should we have a test that verifies that the produced JS package actually works?

hamishmack:
It does run some tests here https://github.com/IntersectMBO/cardano-addresses/actions/runs/9476515970/job/26109526344?pr=254#step:6:1

They are defined in [api.test.ts](https://github.com/IntersectMBO/cardano-addresses/blob/a030bd16eb649376de149116cf91602daa9b58d3/jsapi/test/api.test.ts)

angerman (approved):
Changes look sensible.

PR #256, opened 2024-06-12
Fix windows utf-8 issues

angerman:
This ensure we force the rigth codepages on windows to support utf8 caracters in the haskell source for the tests.


hamishmack:
This has fixed the cross compiled Windows issue.  It does not fix the windows ghc 9.6.5 issue in the GitHub action (so we have left that disabled for now).

hamishmack (approved):

Issue #264, opened 2024-07-13
Web demo is broken: `runCardanoAddressesApi is not defined`


Unisay:
![image](https://github.com/user-attachments/assets/c0feaf77-19eb-4ba8-b44e-366821be9cf2)

paweljakubas:
hi @Unisay 

My strong view is that it is something on application side. Prove:
```
$ echo addr_test1vqq934tq2lnxp9urlltl0vj3sejl92dznu82ngzdj8g4fpc7vnz95 | cabal run cardano-address address inspect
/nix/store/q3yjgay463jbvcnk751pk4r95a31wm47-cabal-install-exe-cabal-3.10.3.0/bin/cabal --project-file=/home/pawel/Work/IntersectMBO/cardano-addresses/.nix-shell-cabal.project run cardano-address address inspect
{
    "address_style": "Shelley",
    "address_type": 6,
    "network_tag": 0,
    "spending_key_hash": "0058d56057e6609783ffd7f7b2518665f2a9a29f0ea9a04d91d15487",
    "spending_key_hash_bech32": "addr_vkh1qpvd2czhuesf0qll6lmmy5vxvhe2ng5lp656qnv3692gwcv4s9j",
    "stake_reference": "none"
}
```

Unisay:
So I looked at the web demo app and noticed the following error in the dev console:
```
Loading module from “http://localhost:3000/dist/cardano-addresses-jsapi.esm.js” was blocked because of a disallowed MIME type (“text/html”)
```

The error went away after I added this fix to the `index.js`:
```
app.use((req, res, next) => {
  res.setHeader('X-Content-Type-Options', 'nosniff');
}); 
```

Unfortunately this isn't enough and the app looks like this:
![image](https://github.com/user-attachments/assets/9b63289b-5a7e-48c5-8d36-a305433b64f3)

It is stuck in the "Loading.." state right from the start, no errors in the dev console, but also no results after clicking the "Inspect" button.

TLDR: the demo app is broken.

abailly:
Removing Javascript support in 4.0.0

PR #267, opened 2024-07-24
Fix JS Demo

Crypto2099:
This is not ready for merging yet but seeks to address Issue #264.

Troubleshooting identified the problem as some duplicated functions in the `cardano-addresses-jsapi.esm.js` file due to legacy `crypto-wrapper.js` and `crc32.js` which are correctly being pulled in via `emcc` now.

There are still a few duplicated functions that need to be traced.


luite:
I'd like to build this PR locally to check the build result, but I'm getting the same error as the CI: running `$~/cardano-addresses/jsapi $ nix develop ..#cardano-addresses-js-shell --command npm install`

https://github.com/IntersectMBO/cardano-addresses/actions/runs/10085769443/job/27887169614?pr=267

What should I do instead?

Crypto2099:
> I'd like to build this PR locally to check the build result, but I'm getting the same npm ERESOLVE error as the CI: running `~/cardano-addresses/jsapi $ nix develop ..#cardano-addresses-js-shell --command npm install`
> 
> https://github.com/IntersectMBO/cardano-addresses/actions/runs/10085769443/job/27887169614?pr=267
> 
> What should I do instead?

This build error I believe was due to my local package and lock getting accidentally updated and included in the commit. I've rolled these changes back this morning. For local testing I am making changes and running the same steps as the `typescript.yml` workflow file in order to build a new copy of the demo.

The "problem" is that the generated file: `cardano-addresses-jsapi.esm.js` that is created by a combination of `./jsbits/emscripten/build.sh` and `./nix/cardano-addresses-js.nix` is creating duplicate function definitions. I went through and manually modified/commented out my local file and identified that all of the functions that were previously being included from `./jsbits/emscripten/crc32.js` and `./jsbits/emscripten/crypto-wrapper.js` were being duplicated (mostly crc32 and cryptonite hashing algorithms that seem to now be getting properly included in the emscripten output from the c headers).

However, there is still 3 lingering functions: 
* `function h$geteuid()`: Lines 25147 & 28089 of output
* `function h$sysconf()`: Lines 25152 & 28094 of output
* `function h$getpwuid_r`: Lines 25157 & 28099 of output
* `function h$foundation_sysrandom_linux`: Lines 24639 & 25170 of output

Crypto2099:
@luite I found at least the first three of the "duplicated" functions included as part of the GHCJS library from this shim file:
https://github.com/ghcjs/ghcjs/blob/b7711fbca7c3f43a61f1dba526e6f2a2656ef44c/lib/boot/shims/pkg/unix.js.pp#L13

Although, I am uncertain where the other (non-commented as stub) versions of these same functions are being included.

luite:
> This build error I believe was due to my local package and lock getting accidentally updated and included in the commit. I've rolled these changes back this morning. For local testing I am making changes and running the same steps as the `typescript.yml` workflow file in order to build a new copy of the demo.

Thanks, I'll try again and have a look at the build artifacts!

Looks like you've found the issues, we've merged some of the JS files into the packages, so they can be removed from the nix build.

GHC 9.6 doesn't automatically build/include `c-sources` with emscripten yet, but a later GHC (I think 9.10) does do that, so a few more custom build steps can be removed by then.

luite:
Thanls it builds here now and can run the tests! Just some still failing because `h$withCBufferOnHeap` is still missing.

luite:
I've pushed fixes for `digest` and `foundation` to fix the missing memory functions. It now fails with `_crc32` missing, since that should still be included by hand (until upgrading to GHC 9.10)

luite:
Use the fixes by updating `cabal.project`:

```
repository ghcjs-overlay
  url: https://raw.githubusercontent.com/input-output-hk/hackage-overlay-ghcjs/acbe0444181da7582127ca8b60eb122cdad1c458
  secure: True
  root-keys:
  key-threshold: 0
  --sha256: sha256-f8LXqyiSlCfi1gbHN3BEB489NCfcC1nut43lyQ0RG1s=
```

Issue #225, opened 2023-05-31
NPM package basic sample does not resolve.


ggcaponetto:
By running the following snippet, I expect the promise to be resolved and to get exit code 0. The process outputs the `info` and hangs instead.

```javascript
const cardanoAddresses = require('cardano-addresses')
const addr = 'addr1gqtnpvdhqrtpd4g424fcaq7k0ufuzyadt7djygf8qdyzevuph3wczvf2dwyx5u'

cardanoAddresses.inspectAddress(addr)
    .then(info => {
        console.log("got info:", info);
    }).catch(e => {
    console.log("something wrong happened:", e);
});
```
Any advice?

abailly:
Removing Javascript support in 4.0.0

PR #279, opened 2024-11-13
Bump version to 4.0.0

palas:


paweljakubas (approved):
Don't we want to update also copyright?
Like : https://github.com/IntersectMBO/cardano-addresses/blob/master/core/cardano-addresses.cabal#L12 ?
etc.

And resolve also https://github.com/IntersectMBO/cardano-addresses/issues/222
and https://github.com/IntersectMBO/cardano-addresses/issues/225 ?

Crypto2099:
I am working towards #225 so hopefully we'll have that before we are fully ready for release. I asked about the copyright issue a while back and did not get a clear answer but definitely agree that it should probably be updated given that the repo is now under Intersect ownership/maintenance.

So, still some work to be done but at least we have a PR to start working against for tagging this as a proper release

ChangeLog.md
```diff
@@ -1,9 +1,12 @@
...
+
+
+## [4.0.0] - 2024-11-13
 
 ### Added
 
+- Simplify filenames by using a hashed prefix instead of the full mnemonic, enhancing portability).
 - `hashKey` for Shelley style
-- Key derivation support for DRep, CCCold and CCHot in accordance to [CIP--105](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0105). Also supported in CLI.
+- Key derivation support for DRep, CCCold and CCHot in accordance to [CIP-0105](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0105). Also supported in CLI.
```
Crypto2099, line 9:
```suggestion
- Key derivation support for DRep, CCCold and CCHot in accordance with [CIP-0105](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0105). Also supported in CLI.
```

abailly:
Just curious: why bumping major version to 4.0.0? I don't see any breaking changes here, but maybe I am wrong?

palas:
@abailly, I felt it was appropriate because there wasn't any release in 2 years, so it seemed quite a big change. But it doesn't have to be a major version. There may end up being breaking changes though, not sure how it is going to unfold.

abailly:
I don't have any fetichism about numbers so if 4.0.0 sounds cool, 4.0.0 it is :) We are planning to do such a release soonish with @paweljakubas

abailly:
So the plan is to do the release once a bunch of changes had been incorporated:
* #288 
* #290

Issue #285, opened 2024-11-27
Compling with later versions of GHC


erikd:
# Context

I have just got this package compiling with `ghc-9.8` and got that PR merged. Most other things in the Cardano/Haskell stack already compiles with `ghc-9.10`. When I started trying to make this package compile with `ghc-9.10` I ran into a number of sticky issues.

* `cardano-addresses` depends on `hjsonschema` which is deprecated.
* `hjsonschema` depends on `protolude` but I have removed that as a dependency from the rest of the tree, because `protolude` was poorly maintained.

What should we do about this? @Crypto2099 @paweljakubas .

erikd:
I think the aim is to compile with a wide range of compilers. Newer compilers tend to become standard compilers over time.

`protolude` is just an alternative `Prelude` and was pretty easy to remove the rest of our code base (involved cargo culting a small amount of functionality into `cardano-base`).

erikd:
> FWIW I have tried removing the dependency to `hjsonschema` and compile the code with 9.8.2 and 9.10.1 and it worked fine.

Is that something that we should consider and merge into `master`?

abailly:
> 
> Is that something that we should consider and merge into `master`?

Would love to but I am a newcomer around here, so not sure what purpose served `hjsonschema`. Perhaps @paweljakubas or @Crypto2099 or @disassembler  would know better?

erikd:
@abailly I have spoken to @disassembler in Slack and as long as removing `hjsonschema` does not break compiling with `ghc-8.10.7` this should be ok to remove.

Crypto2099:
Given that `8.10.7` is the lowest version we currently require via the CI/CD pipeline, if removing this (and other) dependencies does not break the build I would be in favor of removing them.

PR #286, opened 2024-12-12
Remove dependency on hjsonschema

abailly:
This is needed for #285


erikd (changes requested):
@abailly Looks good, but would like to see a minimal diff. There seem to have been a bunch un-needed of formatting changes in this PR.

abailly:
My bad, I have autoformatting on with fourmolu.

erikd (approved):
Looking good. Thanks.

erikd:
@paweljakubas Would appreciate it if you could also review.

paweljakubas (approved):
looks reasonable and good to me

PR #288, opened 2024-12-16
CIP-0129 support

paweljakubas:
PR adds CIP-0129 support

Basically drep, cc hot and cc cold key hashes are decorated with prepended byte. The same is for script hashes.
It affects data payload that is used as hashes. For the rest key hashes and script hashes it is 28-byte payload, here 29-byte payload. 

fix https://github.com/IntersectMBO/cardano-addresses/issues/284

There is going to be additional PR enhancing https://github.com/cardano-foundation/CIPs/tree/master/CIP-0105 test vectors and deprecation comments

In the end we are going to have the following behavior

```
At this moment we support credentials
X : payload is specific byte + 28-byte key hash
X : payload is specific byte+ 28-byte script hash

We are going to support additionally
X: payload is 28-byte key hash (deprecated)
X_script: payload is 28-byte script hash
X_vkh : payload is 28-byte key hash

where
X={drep, cc_cold, cc_hot}
```


abailly:
@paweljakubas If you replace `Tackles https://github.com/IntersectMBO/cardano-addresses/issues/284` with `Fix https://github.com/IntersectMBO/cardano-addresses/issues/284` in the description of your PR, the issue will be automatically closed when the PR is merged.

command-line/test/Command/Script/HashSpec.hs
```diff
@@ -44,85 +44,85 @@ spec = do
         specScriptHashProper "script1w8469gq5ed7xtyf2tqdng5yn7ykgckkfcl38xre8hk3ejk2lcwt"
             [iii|#{verKeyH4}|]
 
-        specScriptHashProper "drep_script13hu7a632tuntrlyjkazzczrzzt7dx6e8v3rc25jd9zcgv68nx9r"
+        specScriptHashProper "drep1ydv8xcdxwpy7zrf805yenpz7cdnap4c87v7zewvmdjyyygq44jy4j"
```
abailly, line 47: Looking at those strings makes me wish those were in a text file usable externally and loaded by our tests so as to be relatively simple to edit and publish

core/lib/Cardano/Address/Script.hs
```diff
@@ -222,6 +227,101 @@ scriptHashFromBytes bytes
...
+        Representative ->
+            let fstByte = 0b00100011 :: Word8
+            in BS.cons fstByte payload
+        CommitteeCold ->
+            let fstByte = 0b00010011 :: Word8
+            in BS.cons fstByte payload
+        CommitteeHot ->
+            let fstByte = 0b00000011 :: Word8
+            in BS.cons fstByte payload
+        _ -> payload
```
abailly, line 265:
```suggestion
    appendByte payload =  maybe payload (`BS.cons` payload) bytePrefix
    bytePrefix = case cred of
        Representative -> Just 0b00100011 
        CommitteeCold -> Just 0b00010011
        CommitteeHot -> Just 0b00000011
        _ -> Nothing
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -222,6 +227,101 @@ scriptHashFromBytes bytes
...
+            maybeToRight ErrScriptHashFromTextWrongDataPart (Bech32.dataPartToBytes dp)
+                >>= maybeToRight ErrScriptHashFromTextWrongHrp . convertBytes hrp
+                >>= maybeToRight ErrScriptHashFromTextWrongPayload . scriptHashFromBytes
+ where
+    convertBytes hrp bytes
+        | hrp == CIP5.drep && checkBSLength bytes 29 =
+              let (fstByte, payload) = first BS.head $ BS.splitAt 1 bytes
+              --   drep          0010....
+              --   scripthash    ....0011
+              in if fstByte == 0b00100011 then
```
abailly, line 297: that's where we could reuse `toBytePrefix/fromBytePrefix` perhaps?

core/lib/Cardano/Address/Script.hs
```diff
@@ -222,6 +227,101 @@ scriptHashFromBytes bytes
...
+        Left _ -> do
+            (hrp, dp) <- first (const ErrScriptHashFromTextInvalidString) $
+                Bech32.decodeLenient txt
+
+            maybeToRight ErrScriptHashFromTextWrongDataPart (Bech32.dataPartToBytes dp)
+                >>= maybeToRight ErrScriptHashFromTextWrongHrp . convertBytes hrp
+                >>= maybeToRight ErrScriptHashFromTextWrongPayload . scriptHashFromBytes
+ where
+    convertBytes hrp bytes
+        | hrp == CIP5.drep && checkBSLength bytes 29 =
```
abailly, line 293: Lots of duplication in there :)

core/lib/Cardano/Address/Script.hs
```diff
@@ -263,14 +363,36 @@ keyHashToText (KeyHash cred keyHash) = case cred of
...
+--   one byte is prepended to vkh only in governance context. The rules how to contruct it are summarized
+--   below
+--
+--   drep       0010....
+--   hot        0000....    key type
+--   cold       0001....
+--
+--   keyhash    ....0010
+keyHashAppendByteCIP0129 :: ByteString -> KeyRole -> ByteString
+keyHashAppendByteCIP0129 payload = \case
```
abailly, line 384: This duplicates the `appendByte` function above, factor out?

core/lib/Cardano/Address/Script.hs
```diff
@@ -408,7 +546,34 @@ prettyErrKeyHashFromText = \case
     ErrKeyHashFromTextWrongDataPart ->
         "Verification key hash is Bech32-encoded but has an invalid data part."
     ErrKeyHashFromTextInvalidHex ->
-        "Invalid hex-encoded string: must be either 28, 32 or 64 bytes"
+        "Invalid hex-encoded string: must be either 28, 32 or 64 bytes."
+
+-- Possible errors when deserializing a script hash from text.
+--
+-- @since 4.0.0
+data ErrScriptHashFromText
```
abailly, line 554: I think it would be useful to provide more details to the user when something fails, like the invalid string, or the wrong prefix, etc.

abailly (commented):
That provides a very good coverage of the features, I would like to suggest we remove some duplication in order to improve the maintainability of this code over time

core/lib/Cardano/Address/Script.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: details are in `prettyErrScriptHashFromText :: ErrScriptHashFromText -> String` function.

paweljakubas: I might tackle this, also `ErrKeyHashFromText` refactoring, as a subtask of https://github.com/IntersectMBO/cardano-addresses/issues/289

README.md
```diff
@@ -592,8 +592,59 @@ drep_xsk1vpdsm49smzmdwhd4kjmm2mdyljjysm746rafjr7r8kgfanj849psw8pfm305g59wng0akw3
...
+
+$ cat drep.vkh
+drep_vkh1sp5xhvmj0asztqfsjyta3cwvq7jppc2rwmfcsggp62va5hqgpg4
+$ cat drep.deprecated
+drep1sp5xhvmj0asztqfsjyta3cwvq7jppc2rwmfcsggp62va538nup0
+$ cat drep.credential
+drep1y2qxs6anwflkqfvpxzg30k8pesr6gy8pgdmd8zppq8ffnksapjznm
+
+$ cardano-address script hash "all [$(cat drep.xvk), active_from 100, active_until 120]"
+drep_script1eeccqnkak63vtp32l3epv5hcn8qpc2nxz4drxzadvj9q78ysk3q
```
Ryun1, line 633: as `drep_script` is the old CIP105, should it be noted as depreciated?

README.md
```diff
@@ -611,8 +662,59 @@ cc_cold_xsk1fp4megtpn4vu4cug2lmsyhg4xvnnar55q6k8wp5e6f2h8jz849ph8v8jhm0qffw8v6ut
...
+
+<details>
+  <summary>How to create script and its hash from cold committee keys (<strong>drep</strong>)</summary>
+
+```console
+
+$ cat cold.vkh
+cc_cold_vkh1d7yw362prvnae5fc8063xdeapws9ptzdgjkqd4dk3qddctn5rch
+$ cat cold.deprecated
+cold1d7yw362prvnae5fc8063xdeapws9ptzdgjkqd4dk3qddcj20wc6
```
Ryun1, line 698: This is the wrong prefix, for CIP105 it should be `cc_cold`

so this should be 
`cc_cold1d7yw362prvnae5fc8063xdeapws9ptzdgjkqd4dk3qddccyzfjm`

README.md
```diff
@@ -611,8 +662,59 @@ cc_cold_xsk1fp4megtpn4vu4cug2lmsyhg4xvnnar55q6k8wp5e6f2h8jz849ph8v8jhm0qffw8v6ut
...
+$ cat cold.credential
+cc_cold1zfhc368fgydj0hx38qal2yeh8596q59vf4z2cpk4k6yp4hqy3mpsx
+
+$ cardano-addcardano-address script hash "all [$(cat cold.xvk), active_from 100, active_until 120]"
```
Ryun1, line 702:
```suggestion
$ cardano-address script hash "all [$(cat cold.xvk), active_from 100, active_until 120]"
```
additional `cardano-address`

README.md
```diff
@@ -611,8 +662,59 @@ cc_cold_xsk1fp4megtpn4vu4cug2lmsyhg4xvnnar55q6k8wp5e6f2h8jz849ph8v8jhm0qffw8v6ut
...
+cc_cold_script18zy6g0vu7ajzmzamkvysfzc0nnfpf8w3n7404xxhaz2jqexhzw5
+$ cardano-addcardano-address script hash "all [$(cat cold.deprecated), active_from 100, active_until 120]"
```
Ryun1, line 704:
```suggestion
$ cardano-address script hash "all [$(cat cold.deprecated), active_from 100, active_until 120]"
```

additional `cardano-address`

README.md
```diff
@@ -611,8 +662,59 @@ cc_cold_xsk1fp4megtpn4vu4cug2lmsyhg4xvnnar55q6k8wp5e6f2h8jz849ph8v8jhm0qffw8v6ut
...
+
+$ cat cold.vkh
+cc_cold_vkh1d7yw362prvnae5fc8063xdeapws9ptzdgjkqd4dk3qddctn5rch
+$ cat cold.deprecated
+cold1d7yw362prvnae5fc8063xdeapws9ptzdgjkqd4dk3qddcj20wc6
+$ cat cold.credential
+cc_cold1zfhc368fgydj0hx38qal2yeh8596q59vf4z2cpk4k6yp4hqy3mpsx
+
+$ cardano-addcardano-address script hash "all [$(cat cold.xvk), active_from 100, active_until 120]"
+cc_cold_script18zy6g0vu7ajzmzamkvysfzc0nnfpf8w3n7404xxhaz2jqexhzw5
```
Ryun1, line 703: again, this is the CIP105 representation, I think it should be marked as depreciated

README.md
paweljakubas: good catch, thanks, fixed

README.md
paweljakubas: good catch, thanks, fixed

README.md
paweljakubas: good catch, thanks, fixed

README.md
paweljakubas: IMHO it is correct as it is. To sum up, let's assume x is one of `drep`, `cc_cold` and `cc_hot`. We have the following:
1. `x_xsk` - extended signing key
2. `x_sk` - signing key
3. `x_xvk` - extended verification key
4. `x_vk` - verfication key
5a. `x_vkh` - verification key  hash - here 28-byte payload - it was added recently
5b. `x_script` - script hash - here 28-byte payload - it was there already
6. `x` - verifcation key hash - here 28-byte payload - it is DEPRECATED, we aim to use 5a now instead, ie., `x_vkh`
7. `x` credential which is 28-byte payload plus prepended byte, so overall 29-byte payload. Please notice that we can express here `x_xvk`, `x_script` for all `x`. It overlaps with deprecated 6, but fortunatley it is 29-byte so we can differentiate. 
All is in accordance to CIP-0129, CIP-0105 but also with https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005 (see `cc_cold_script` , `cc_hot_script` and `drep_script`). Hope it makes sense. At first glance I also thought it is erroneous, but now I would argue it is correct as we have.

core/lib/Cardano/Address/Script.hs
```diff
@@ -222,6 +227,121 @@ scriptHashFromBytes bytes
     | BS.length bytes /= credentialHashSize = Nothing
     | otherwise = Just $ ScriptHash bytes
 
+-- | Encode a 'ScriptHash' to bech32 'Text' or hex if key role is unknown.
+--  If one wants to include, valid in governance roles only, additional byte
+--  as specified in CIP-0129, the function needs to be called with withByte=true.
```
abailly, line 232:
```suggestion
--  In the case of governance role, if one wants to include additional byte
--  as specified in [CIP-0129](https://github.com/cardano-foundation/CIPs/blob/master/CIP-0129/README.md), the function needs to be called with withByte=true.
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -222,6 +227,121 @@ scriptHashFromBytes bytes
...
+        if withByte then
+            T.decodeUtf8 $ encode (EBech32 CIP5.cc_hot) $ appendByte scriptHash
+        else
+            T.decodeUtf8 $ encode (EBech32 CIP5.cc_hot_script) scriptHash
+    Unknown ->
+        T.decodeUtf8 $ encode EBase16 scriptHash
+    _ ->
+        T.decodeUtf8 $ encode (EBech32 CIP5.script) scriptHash
+  where
+-- | In accordance to CIP-0129 (https://github.com/cardano-foundation/CIPs/tree/master/CIP-0129)
```
abailly, line 257: I would move this comment into the haddock comment of the function itself.

core/lib/Cardano/Address/Script.hs
```diff
@@ -222,6 +227,121 @@ scriptHashFromBytes bytes
     | BS.length bytes /= credentialHashSize = Nothing
     | otherwise = Just $ ScriptHash bytes
 
+-- | Encode a 'ScriptHash' to bech32 'Text' or hex if key role is unknown.
+--  If one wants to include, valid in governance roles only, additional byte
+--  as specified in CIP-0129, the function needs to be called with withByte=true.
+--
+-- @since 4.0.0
+scriptHashToText :: ScriptHash -> KeyRole -> Bool -> Text
```
abailly, line 235: this `Bool` parameter feels somewhat wrong, ie. [boolean blindness](https://existentialtype.wordpress.com/2011/03/15/boolean-blindness/). Would it not be better to include this extra byte into the `KeyRole`? e.g 

```
data KeyRoles = Bare KeyRole | WithByte KeyRole
```

I am not saying we should do it here, perhaps would it be useful to discuss the design first in a follow-up PR?

core/lib/Cardano/Address/Script.hs
```diff
@@ -252,25 +372,59 @@ keyHashFromBytes (cred, bytes)
     | otherwise = Just $ KeyHash cred bytes
 
 -- | Encode a 'KeyHash' to bech32 'Text' or hex is key role unknown.
+--  If one wants to include, valid in governance roles only, additional byte
+--  as specified in CIP-0129, the function needs to be called with withByte=true.
 --
 -- @since 3.0.0
-keyHashToText :: KeyHash -> Text
-keyHashToText (KeyHash cred keyHash) = case cred of
+keyHashToText :: KeyHash -> Bool -> Text
```
abailly, line 379: Same remark as above, which makes me think we need something more abstract like

```
data WithByte a = WithByte a | Bare a 
```

core/lib/Cardano/Address/Script.hs
```diff
@@ -252,25 +372,59 @@ keyHashFromBytes (cred, bytes)
...
+--   drep       0010....
+--   hot        0000....    key type
+--   cold       0001....
+--
+--   keyhash    ....0010
+--   This is on top of X_vkh, where X={drep, cc_hot, cc_hot}, which lacks the additional byte.
+--   In `keyHashFromText` we additionally
+--   support reading legacy X which also lacks the additional byte, and has the same payload as
+--   as the corresponding X_vkh.
+keyHashAppendByteCIP0129 :: ByteString -> KeyRole -> ByteString
```
abailly, line 418: This is the same comment than in the `scriptToHashText` function, perhaps just a link would suffice?

abailly (approved):
Some minor adjustments which might be done in a later PR. I think it would be worthwhile to have a more involved discussion about the design of keyrole/keyhash to have a better and safer handling of CIP-129 extensions

README.md
Ryun1: > x_script - script hash - here 28-byte payload - it was there already

I believe the CIP105 `_script` have been depreciated
because CIP129's header allows differentiation between script-based and key-based whilst using the same prefix, so the `_script are not needed to differentiate.

Although reading CIP-0005 currently I believe there is a misleading mistake with this.
Because the `_script` prefixes are contained within both the [Hashes table](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005#hashes) AND the [Depreciated Governance Prefixes table](https://github.com/cardano-foundation/CIPs/tree/master/CIP-0005#deprecated-governance-prefixes).

I think it is a mistake that the `_script` are in both tables.
I have made a [PR to CIP-05 to fix this](https://github.com/cardano-foundation/CIPs/pull/965) (and the rendering) as well as reached out to the CIP129 author to confirm itension.

paweljakubas: we have the following (let's take x=drep):
1. `drep_vkh` - 28-byte hash verification key hash introduced in CIP105
2. this guy can participate in defining script, for example `all [ drep_vkh, after slot1]`
3. The script in point 2 can be hashed, and we have 28-byte script hash with hrp `drep_script`
4. CIP105 introduced `drep` having 28-byte plus 1 byte prepending which has information about either it is drep/cc_cold/cc_hot and if it is script/verification key hash. So it has 29-byte !!!!
5. What is deprecated is using `drep`  as point 1, so verification key hash. Now we use `drep_vkh` but after adding 1 more byte we can also express it as 29-byte `drep`. 

@Ryun1 Please notice that `X_script` is in accordance to other prefixes that have script hashes. 
Of course I will adopt any change decided;-) I am just soldier here.

paweljakubas: Overall, PR as it stands support legacy treatment and new way. Please refer to Readme.md sections:
1. How to generate drep keys
2. How to create script and its hash from drep keys
3. How to generate cold committee keys
4. How to create script and its hash from cold committee keys
5. How to generate hot committee keys
6. How to create script and its hash from hot committee keys

So, besides interesting discussion, we have all is needed to support fully new treatment of governance keys. We also have backward compatibility, for example reading 28-byte `drep`.

README.md
```diff
@@ -592,8 +592,59 @@ drep_xsk1vpdsm49smzmdwhd4kjmm2mdyljjysm746rafjr7r8kgfanj849psw8pfm305g59wng0akw3
...
+drep1sp5xhvmj0asztqfsjyta3cwvq7jppc2rwmfcsggp62va538nup0
+
+$ cat drep.deprecated | bech32
+80686bb3727f602581309117d8e1cc07a410e14376d3882101d299da
+
+$ cardano-address key hash --with-byte < drep.xvk > drep.credential
+drep1y2qxs6anwflkqfvpxzg30k8pesr6gy8pgdmd8zppq8ffnksapjznm
+
+$ cat drep.credential | bech32
+2280686bb3727f602581309117d8e1cc07a410e14376d3882101d299da
```
Crypto2099, line 614: > $ cardano-address key hash --with-byte < drep.xvk > drep.credential
drep1y2qxs6anwflkqfvpxzg30k8pesr6gy8pgdmd8zppq8ffnksapjznm

Apologies for being a bit late to the game here, but the `CIP-105` version is deprecated and the `CIP-129` version should be considered default. Does it make more sense to move the optional function modifier flag (in this case `--with-byte`) to instead be present to trigger the "Legacy" or "Deprecated" mode and return the `CIP-105` version?

Crypto2099: Example, instead of requiring the `--with-byte` flag to generate the most-current version of the expected output, we could instead add a `--cip-105` or `--legacy` flag to generate the deprecated model?

README.md
Crypto2099: See my comment in my own review about potentially changing to default to returning CIP-0129 format and providing an optional flag to explicitly force CIP-105 deprecated formats

README.md
```diff
@@ -611,8 +662,59 @@ cc_cold_xsk1fp4megtpn4vu4cug2lmsyhg4xvnnar55q6k8wp5e6f2h8jz849ph8v8jhm0qffw8v6ut
 $ cardano-address key public --with-chain-code < cold.xsk > cold.xvk
 cc_cold_xvk1dg8d5du0v4ukqkfgset50xncudhwlfzz2p6epv096x0ndl8jsgzzdqzpe9yw37u7mu7xaenhv7242990ps3sn8jcg52yx7n3fuke9kst5t2py
 
-$ cardano-address key hash < cold.xvk > cold
+$ cardano-address key public --without-chain-code < cold.xsk > cold.vk
+cc_cold_vk1dg8d5du0v4ukqkfgset50xncudhwlfzz2p6epv096x0ndl8jsgzqmwj2x5
```
Crypto2099, line 666: I notice that we have `--with-chain-code` and `--without-chain-code` flags here, what is the default behavior if no chain code is specified?

Crypto2099 (approved):
My comments are mostly related to design patterns and I would hope that we address them or make a decision prior to moving forward so we don't establish precedent that can then be used as an appeal to history to keep it the same going forward. But I'm okay with merging these changes as-is as well to advance the project forward.

core/lib/Cardano/Address/Script.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: done

core/lib/Cardano/Address/Script.hs
paweljakubas: I will address that in https://github.com/IntersectMBO/cardano-addresses/issues/289

core/lib/Cardano/Address/Script.hs
paweljakubas: I will address that in forthcoming refactoring as part of https://github.com/IntersectMBO/cardano-addresses/issues/289

README.md
paweljakubas: good suggestion. As mentioned above we have `--cip-0105` flag now on demand

README.md
paweljakubas: there is error if one does not specify this. One needs to explicitly state either `--with-chain` or `-without-chain`
https://github.com/IntersectMBO/cardano-addresses/blob/master/command-line/lib/Options/Applicative/Public.hs#L31

PR #291, opened 2025-01-13
Scrap hex and rely on bech32 for converting to hex

paweljakubas:
fix https://github.com/IntersectMBO/cardano-addresses/issues/282


command-line/lib/Command/Key/Hash.hs
```diff
@@ -37,30 +34,26 @@ import System.IO.Extra
 import qualified Cardano.Codec.Bech32.Prefixes as CIP5
 import qualified Data.ByteString as BS
 
-newtype Cmd = Hash
-    { outputFormat :: FormatType
-    } deriving (Show)
+data Cmd = Hash
```
abailly, line 37: This feels a bit odd, don't we have other sub-commands for `key` command?

abailly (approved):
Love that we are _removing_ stuff instead of adding more and leveraging other tools! 
Thank you for that. Just some minor comment about whether or not we could avoid creating a specific type for the `hash` sub-command but feel free to ignore

PR #292, opened 2025-01-15
Get rid of js

paweljakubas:
fix https://github.com/IntersectMBO/cardano-addresses/issues/290


abailly (changes requested):
I think we should update the documentation to point at alternative solutions, eg. meshjs, blaze, others?

abailly (approved):
Awesome! ❤️

PR #295, opened 2025-01-30
Unify core and command-line into one package

paweljakubas:


abailly (commented):
How about having everything at the top-level, eg. have a `.cabal` file at top-level with `src/` and `test/`? eg. move everything under `command-line/*` to toplevel

abailly:
Also I would just have a single package called `cardano-address` not `cardano-address-cli`

abailly:
Here is how the package looks like currently on hackage: https://hackage.haskell.org/package/cardano-addresses-4.0.0/candidate

abailly (approved):
I think this is both much simpler and much clearer ❤️

Issue #294, opened 2025-01-24
Release new version (4.0.0)


abailly:
- [x] Create PR bumping versions from 3.12 -> 4.0.0 in files. Also clean Copyright, add missing ones, clean README.md
        -> https://github.com/IntersectMBO/cardano-addresses/pull/279
- [x] Unify core and command packaged under one package
       -> https://github.com/IntersectMBO/cardano-addresses/pull/295
- [x]  Release in repo by adding tag, attaching needed artifacts, etc. and add GH action file handling this
       -> https://github.com/IntersectMBO/cardano-addresses/pull/297
       -> https://github.com/IntersectMBO/cardano-addresses/releases/tag/4.0.0
- [x] Create hackage account for CF
- [x] Publish cardano-crypto to hackage
       -> https://hackage.haskell.org/package/cardano-crypto-1.1.3/candidate
- [x] Publish cardano-address to hackage along with GH action file handling this
       -> https://hackage.haskell.org/package/cardano-addresses-4.0.0

paweljakubas:
PR addressing first task -> https://github.com/IntersectMBO/cardano-addresses/pull/279

abailly:
I think we might want also to merge `core/` and `command-line/` into a single package: Seems to me there's no reason anymore to separate the two, and it will simplify publication workflow. WDYT @paweljakubas ?

paweljakubas:
good idea, I will take care of it 👍

abailly:
If we want to publish the package to hackage we need to provide bounds in the cabal file

PR #296, opened 2025-01-31
Fix maintainer address

abailly:
Contribute to #294


paweljakubas (approved):
:+1:  I was thinking about this yesterday and was not sure how to proceed here given what is written in copyright.

abailly:
I think this is different: We want to provide an up-to-date contact point should people want to reach out, and for all practical purpose this is the best contact point we can have.

PR #301, opened 2025-02-20
Set package bounds

paweljakubas:


abailly:
Why are the `ci/` jobs failing whereas the GHA are succeeding with both GHC versions?

abailly:
I think this is a similar problem than the one I experienced when I tried to publish cardano-wallet packages to CHaP: There's something problematic with GHC 8.10.7 in Haskell.nix or whatever that prevents building packages which depend on doctest. Perhaps @angerman could confirm?

angerman:
@abailly they all failed for cabal-doctest, @paweljakubas seems to have fixed them! 🎉

erikd (approved):

PR #302, opened 2025-02-26
Prepare publication to hackage

abailly:
* put bounds on cardano-crypto dependency: we set the bound in such a way both CHaP and Hackage can provide a valid version
* replace cryptonite with crypton

#294

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -10,10 +10,6 @@ repository cardano-haskell-packages
     c00aae8461a256275598500ea0e187588c35a5d5d7454fb57eac18d9edb86a56
     d4a35cd3121aa00d18544bb0ac01c3e1691d618f462c46129271bccf39f7e8ee
 
-index-state:
-  , hackage.haskell.org 2024-11-14T09:17:39Z
-  , cardano-haskell-packages 2024-11-20T20:05:41Z
-
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them
 
@@ -21,6 +17,10 @@ if impl(ghc>=9.8)
   allow-newer: *:template-haskell, *:base, *:deepseq, *:ghc-prim, *:time
 
 active-repositories: hackage.haskell.org, cardano-haskell-packages
+
+index-state:
+  , hackage.haskell.org 2024-11-14T09:17:39Z
+  , cardano-haskell-packages 2025-01-23T00:00:00Z
 
 packages:
   cardano-addresses.cabal
--- a/cardano-addresses.cabal
+++ b/cardano-addresses.cabal
@@ -94,10 +94,10 @@ library
     , bech32-th >= 1.1.7 && < 1.2
     , binary
     , bytestring >= 0.10.6 && < 0.13
-    , cardano-crypto
+    , cardano-crypto >= 1.2.0
     , cborg >= 0.2.1 && <0.3
     , containers >= 0.5 && < 0.8
-    , cryptonite
+    , crypton >= 0.32 && < 1.1
     , deepseq >= 1.4.4.0 && < 1.6
     , digest
     , either
@@ -197,7 +197,7 @@ test-suite unit
     , bytestring >= 0.10.6 && < 0.13
     , cardano-addresses
     , cardano-crypto
-    , cryptonite
+    , crypton
     , containers >= 0.5 && < 0.8
     , hspec >= 2.11.0 && < 2.12
     , hspec-golden >=0.1.0.3 && <0.2
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "CHaP": {
       "flake": false,
       "locked": {
-        "lastModified": 1736937016,
-        "narHash": "sha256-dmLSu2SvSaTDjSE03cU6DwY62J3nWJbVhIn/kKtMwJg=",
+        "lastModified": 1740589390,
+        "narHash": "sha256-3sejZgLk9MkGBWz5p+XH8orjpb1RDwXh4Kn1j9kWjDY=",
         "owner": "intersectmbo",
         "repo": "cardano-haskell-packages",
-        "rev": "045875beec586ff57a7333c0563fd5c2b1a308fa",
+        "rev": "1e58f5e988d05265e8db958fa044b5a440358226",
         "type": "github"
       },
       "original": {


cabal.project
```diff
@@ -11,8 +11,8 @@ repository cardano-haskell-packages
     d4a35cd3121aa00d18544bb0ac01c3e1691d618f462c46129271bccf39f7e8ee
 
 index-state:
-  , hackage.haskell.org 2024-11-14T09:17:39Z
-  , cardano-haskell-packages 2024-11-20T20:05:41Z
+  , hackage.haskell.org 2025-02-26T10:59:23Z
+  , cardano-haskell-packages 2025-02-15T18:39:38Z
```
paweljakubas, line 15: do we really need this now?

paweljakubas, line 15: if no then we can delete lines up to 11

abailly: Yes, because 1.2.0 is not available with the current indices. Note that indices are not used with plain cabal

abailly: Ah wait, you mean: do we need pinning down the versions? Perhaps not, indeed

paweljakubas: ok, next thing: do we really need:
```
active-repositories: hackage.haskell.org, cardano-haskell-packages 
```
?

paweljakubas: I bet when we will have 1.2.0 in the indices when we would be able to delete `cardano-haskell-packages` and also the definition at the beginning of this file, correct?

abailly: My reasoning is that we don't publish `cabal.project` to hackage so whatever we put here is only useful within the context of the CI, it doesn't hurt but has to be consistent.

angerman:
Hi guys, I'm a bit confused what you want to do here? @paweljakubas seems to suggest removing CHaP because there are no dependencies on packages exclusively in CHaP. Remember CHaP exists, so that IOG does not spam the Hackage index. We can release way more frequently to CHaP, and release packages to CHaP as fixups of upstream packages until they are fixed. We can also release to CHaP stuff that should not necessarily live in Hackage. Yet we want `cabal` to "just work" hence the additional repository.

Now if you remove the index states, you basically lose predictablity of the build-plan because any solution will now depend on whatever state your ~/.cabal hackage index state has. Two different people can now get different build plans and different build results (unless you go and commit a freeze file, or some such).

The build-plan is deterministic given the same environment and index-state. The index-state tells cabal up to which point in time to consider data in the index file.

---

From the CI error
```
Resolving dependencies...
make-install-plan: dieVerbatim: user error (Error: make-install-plan: Could not resolve dependencies:
[__0] trying: cardano-addresses-4.0.0 (user goal)
[__1] next goal: cardano-crypto (dependency of cardano-addresses)
[__1] rejecting: cardano-crypto-1.1.2 (conflict: cardano-addresses =>
cardano-crypto>=1.2.0)
[__1] skipping: cardano-crypto-1.1.1, cardano-crypto-1.1.0 (has the same
characteristics that caused the previous version to fail: excluded by
constraint '>=1.2.0' from 'cardano-addresses')
[__1] fail (backjumping, conflict set: cardano-addresses, cardano-crypto)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: cardano-addresses, cardano-crypto
)
```
so we can't find `cardano-crypto >= 1.2` as set in the [cardano-addresses.cabal](https://github.com/IntersectMBO/cardano-addresses/blob/07058911ec97e4d61e5cda09a9527546eae08b6b/cardano-addresses.cabal#L97) file.

### for `cabal`
1. `cardano-crypto` however [does not exist in hackage](https://hackage.haskell.org/package/cardano-crypto). Hence, we can't drop CHaP (unless someone uploads `cardano-crypto` to hackage).
2. `cardano-crypto` in version 1.2.0 [does exist in chap](https://chap.intersectmbo.org/package/cardano-crypto-1.2.0/). With a timestamp of `2025-01-22T23:05:02Z`. Hence the cardano-haskell-packages timestamp needs to be at least `2025-01-22T23:05:02Z`.
### for `nix/hydra`
3. In addition to that we need to ensure that the CHaP reference in the `flake.lock` file that `nix` uses is also at least that new, otherwise, no matter how new the `index-state` is that's set in the `cabal.project` file, `nix` won't be able to see it.
```
• Updated input 'CHaP':
    'github:intersectmbo/cardano-haskell-packages/045875beec586ff57a7333c0563fd5c2b1a308fa?narHash=sha256-dmLSu2SvSaTDjSE03cU6DwY62J3nWJbVhIn/kKtMwJg%3D' (2025-01-15)
  → 'github:intersectmbo/cardano-haskell-packages/1e58f5e988d05265e8db958fa044b5a440358226?narHash=sha256-3sejZgLk9MkGBWz5p%2BXH8orjpb1RDwXh4Kn1j9kWjDY%3D' (2025-02-26)
  ```
  
 ---
 
 This diff fixes both:
 ```
 diff --git a/cabal.project b/cabal.project
index aafceda6..5d915b6c 100644
--- a/cabal.project
+++ b/cabal.project
@@ -10,6 +10,10 @@ repository cardano-haskell-packages
     c00aae8461a256275598500ea0e187588c35a5d5d7454fb57eac18d9edb86a56
     d4a35cd3121aa00d18544bb0ac01c3e1691d618f462c46129271bccf39f7e8ee

+index-state:
+  , hackage.haskell.org 2024-11-14T09:17:39Z
+  , cardano-haskell-packages 2025-01-23T00:00:00Z
+
 -- See CONTRIBUTING for information about these, including some Nix commands
 -- you need to run if you change them

diff --git a/flake.lock b/flake.lock
index ce03c5fd..f3b37597 100644
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "CHaP": {
       "flake": false,
       "locked": {
-        "lastModified": 1736937016,
-        "narHash": "sha256-dmLSu2SvSaTDjSE03cU6DwY62J3nWJbVhIn/kKtMwJg=",
+        "lastModified": 1740589390,
+        "narHash": "sha256-3sejZgLk9MkGBWz5p+XH8orjpb1RDwXh4Kn1j9kWjDY=",
         "owner": "intersectmbo",
         "repo": "cardano-haskell-packages",
-        "rev": "045875beec586ff57a7333c0563fd5c2b1a308fa",
+        "rev": "1e58f5e988d05265e8db958fa044b5a440358226",
         "type": "github"
       },
       "original": {
```

---

Can't I get away without the `index-state` stanzas in the `cabal.project` file? Yes, you can. If you bump the `CHaP` reference in the `flake.lock` past `2025-01-22T23:05:02Z` (e.g. apply only the `flake.lock` part from the patch above), it will also build. I would again, however like to point out that this will not give you a deterministic build-plan. The build-plan is now implicitly dependent on the state of the hackage and CHaP index that cabal sees. It will be deterministic in `nix` due to the pinning of hackageNix and CHaP, but it will not be deterministic anywhere outside of nix, and the nix build in CI and the build outside of nix will now not necessarily choose the same build-plan anymore. As @abailly said, the `cabal.project` file is not committed to hackage anyway, it's not part of the source dist. I'd still encourage you to keep index-states in your `cabal.project` file to be explicit about the build-plan. You are running the risk of otherwise having someone introduce a package to hackage that breaks your project (or worse, compromises it with malicious code); the hackage index is append-only, thus an index-state will freeze it in time.

abailly:
@angerman Thanks a lot for your detailed reply. I don't have time today to go through all this but I can answer your question about the purpose of all this. We want to publish cardano-address package to Hackage for a couple of reasons:

1. this removes one hurdle in the users' path to install and use it as they could now (assuming they have a compatible ghc/cabal toolchain installed) do `cabal install cardano-address` anywhere and have a reasonable chance of being able to run it (ie. no need to define additional repository)
2. Hackage is still _the_ central place for hosting packages, giving us haddock hosting and download statistics for free, and additional visibility
3. Deploying to Hackage is completely automatable, ie. can be put as part of a CI/CD workflow without waiting for someone with merge rights in CHaP to publish it.

So what I am trying to do here is just to trim the `cabal.project` to ensure we can build cardano-addresses with latest dependency versions on `Cardano-crypto` (1.2.0 on CHaP, 1.3.0 on Hackage, versions being different obviously to avoid messing up) with and without CHaP. 

The part I was missing was the flake update, I will take care of this.

abailly:
> Two different people can now get different build plans and different build results (unless you go and commit a freeze file, or some such).

Would you suggest we put _exact_ bounds on the package's dependencies? It seems the current best practice on hackage (or CHaP for that matter) is rather to define intervals which give more flexibility for consumers of libraries to use those? 

Ideally, I think it would make sense to use somewhat laxer constraints on the library and stricter constraint on executable, but is this possible with Cabal?

> Remember CHaP exists, so that IOG does not spam the Hackage index. 

Well, turns out this package is not an IOG package but an Intersect package.

> We can also release to CHaP stuff that should not necessarily live in Hackage. 

`cardano-address` is a widely used package (it ranked 3rd in the list of tools used by people in the latest ecosystem survey) that is Cardano specific but not IOG or Intersect specific, and that does not depend on the whole ecosystem of Haskell Cardano libraries used in the node. It seems to me it makes sense for it to live in Hackage. This was discussed and accepted in December when I presented this plan to the Intersect OSC, which IIRC you attended @angerman. 

> Yet we want cabal to "just work" hence the additional repository.

Exactly, that's my goal too, and I want it to work whether or not one is using nix.

angerman:
I'm not opposed to pushing this to hackage. I was only pointing out that in doing so, its dependencies must also be pushed (e.g. cardano-crypto, not sure if other non-hackage dependencies exist), otherwise it's not buildable (without the end user manually adding CHaP as an extra repo, and then the point becomes moot).

The cabal.project file won't be by up in the hackage sdist, it's purely for _this_ repo and it's development. So index-states are irrelevant for the hackage released package.

Pinning packages by exact version does not protect you from malicious revisions.

angerman:
I'm not opposed to pushing this to hackage. I was only pointing out that in doing so, its dependencies must also be pushed (e.g. cardano-crypto, not sure if other non-hackage dependencies exist), otherwise it's not buildable (without the end user manually adding CHaP as an extra repo, and then the point becomes moot).

The cabal.project file won't be by up in the hackage sdist, it's purely for _this_ repo and it's development. So index-states are irrelevant for the hackage released package.

Pinning packages by exact version does not protect you from malicious revisions.

paweljakubas (approved):
@angerman @abailly thanks for the extensive comments. I am :ok:  when PR is green

PR #303, opened 2025-03-01
Tweak cabal configuration to support CHaP and Hackage

abailly:
As of this commit, version 1.3.0 of cardano-crypto has been published to https://hackage.haskell.org/. This version is identical to version 1.2.0 which is published to https://chap.intersectmbo.org, so it's fine to leave the option to the end-user to use one or the other repository.

We keep the CHaP configuration in the `cabal.project` with updated indices as it can be useful for contributors and provides a reference with "pinned" versions for building in CI, but as it's not published to the package repository, it won't affect build plan downstream which will use whatever local configuration.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -19,7 +19,7 @@ if impl(ghc>=9.8)
 active-repositories: hackage.haskell.org, cardano-haskell-packages
 
 index-state:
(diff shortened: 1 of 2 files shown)


paweljakubas (approved):
:+1:

PR #305, opened 2025-03-28
Polish haddock docs

paweljakubas:
- [x] go through all exposed docs and polish it according to reccommendations
- [x] add byron root address and payment key bidirectional relation docs
- [x] add haddock github run

fix https://github.com/IntersectMBO/cardano-addresses/issues/304


paweljakubas:
@abailly I think we are fine now :+1:

abailly (approved):

abailly:
We should be able to push a revision to hackage to update documents there.

PR #310, opened 2025-10-17
Refactor mnemonics to accommodate mnemonic language

paweljakubas:
The first PR for https://github.com/IntersectMBO/cardano-addresses/issues/309

In this PR I have added Dictionary to function arguments of `mnemonicToText` and `mkMnemonic` in `Cardano.Mnemonic` module. Thanks to that it will be possible to use different language dictionaries in the forthcoming PR to generate mnemonics. All functionality and testing stays the same modulo code adjustments.

Moreover, `MnemonicLanguage` data type was added and parser added to read language optional argument. The default language will be `english` to make the whole solution backward compatible. Finally, `MnemonicLanguage` was added, as optional argument with `english` default to `generate` cmd.

### DIFF
--- a/lib/Cardano/Mnemonic.hs
+++ b/lib/Cardano/Mnemonic.hs
@@ -261,7 +261,7 @@ genEntropy =
 --
 -- prop> mkMnemonicWithDict (mnemonicToTextWithDict mnemonic dictionary) dictionary == Right mnemonic
 --
--- @since 4.1.0
+-- @since 4.0.1
 mkMnemonicWithDict
     :: forall (mw :: Nat) (ent :: Nat) csz.
      ( ConsistentEntropy ent mw csz
@@ -350,7 +350,7 @@ instance (KnownNat csz) => Basement.Exception (MnemonicException csz)
 
 -- | Convert a 'Mnemonic' to a sentence of a specified dictionary mnemonic words.
 --
--- @since 4.1.0
+-- @since 4.0.1
 mnemonicToTextWithDict
     :: Mnemonic mw
     -> Dictionary
--- a/lib/Options/Applicative/MnemonicLanguage.hs
+++ b/lib/Options/Applicative/MnemonicLanguage.hs
@@ -48,15 +48,15 @@ data MnemonicLanguage
     = English
     deriving (Generic, Show, Bounded, Enum, Eq)
 
--- | MnemonicLanguage displays according to [ISO 639-2 Code](https://www.loc.gov/standards/iso639-2/php/code_list.php)
+-- | MnemonicLanguage displays according to two-letter [ISO 639-1 Code](https://www.loc.gov/standards/iso639-2/php/code_list.php)
 mnemonicLanguageToString :: MnemonicLanguage -> String
-mnemonicLanguageToString English = "eng"
+mnemonicLanguageToString English = "en"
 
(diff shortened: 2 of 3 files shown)


lib/Options/Applicative/MnemonicLanguage.hs
```diff
@@ -0,0 +1,78 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE LambdaCase #-}
+
+{-# OPTIONS_HADDOCK hide #-}
+
+-- |
+-- Copyright: 2020 Input Output (Hong Kong) Ltd., 2021-2022 Input Output Global Inc. (IOG), 2023-2025 Intersect
```
abailly: I don't think you need to add any copyright notice date prior to current year.

lib/Options/Applicative/MnemonicLanguage.hs
```diff
@@ -0,0 +1,78 @@
...
+    , showDefaultWith
+    , value
+    )
+
+--
+-- Type
+--
+
+data MnemonicLanguage
+    = English
```
abailly: Shouldn't we use https://www.loc.gov/standards/iso639-2/php/code_list.php ?

lib/Options/Applicative/MnemonicLanguage.hs
paweljakubas: good point -> changed

lib/Command/RecoveryPhrase/Generate.hs
```diff
@@ -30,25 +33,30 @@ import qualified Data.Text as T
...
+data Cmd = Generate
     { size :: MnemonicSize
+    , language :: MnemonicLanguage
     } deriving (Show)
 
 mod :: (Cmd -> parent) -> Mod CommandFields parent
 mod liftCmd = command "generate" $
     info (helper <*> fmap liftCmd parser) $ mempty
-        <> progDesc "Generate an English recovery phrase"
+        <> progDesc "Generate a recovery phrase for a specified mnemonic and language. "
```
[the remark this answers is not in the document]
paweljakubas: done

lib/Cardano/Mnemonic.hs
```diff
@@ -316,17 +348,27 @@ fromUtf8String = T.pack . Basement.toList
...
 --
--- @since 1.0.0
+-- @since 4.1.0
 mnemonicToText
     :: Mnemonic mw
+    -> Dictionary
     -> [Text]
-mnemonicToText =
-    map (fromUtf8String . dictionaryIndexToWord Dictionary.english)
+mnemonicToText mnemonic dictionary =
```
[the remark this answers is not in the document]
paweljakubas: very good point. In the end I decided to leave `mnemonicToText` and `mkMnemonic` like it was before in order not to introduce breaking changes. And introduced new `mnemonicToTextWithDict` and `mkMnemonicWithDict` - of course those more general, with dictionary argument, are used inside the specialized ones, ie. with english dictionary

abailly (approved):
LGTM. I am curious about the version change and therefore the versioning scheme: if we follow PVP then `4.0.0 -> 4.1.0` is a major version change, but it seems we are not adhering to this versioning scheme? 
We should probably clarify that somewhere, perhaps in the `CONTRIBUTING.md`?

PR #313, opened 2025-10-27
Allow optparse-applicative to use 0.19

paweljakubas:
Addresses https://github.com/IntersectMBO/cardano-addresses/issues/311


jasagredo:
Thanks!! can we release this into Hackage now that it is merged?

PR #333, opened 2026-02-23
Fix nix

paweljakubas:
Many fixes of `nix develop` plus `build-docker` inside nix, etc. Plus opulent docs what was added

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -15,17 +15,17 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-   - name: Pull container
-      uses: input-output-hk/actions/devx@latest
-      with:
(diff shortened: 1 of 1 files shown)


nix/haskell.nix
angerman: Do we still need 810?

angerman:
> @angerman any chance to make ci/eval not complaining, meaning green, and start weekend with good feeling? Many thanks in advance ;-)

Hi @paweljakubas there are no hydraJobs in this anymore. That's why hydra is failing? Don't we want to build at least cardano-addresses binaries with hydra? What would be the appropriate build targets that we want to build in CI?

angerman:
@paweljakubas ci/eval should not be required, and also the webhook that triggers hydra builds on ci.iog.io has been disabled. Let me know if you need anything else.

PR #334, opened 2026-02-23
Notorize macos executables

paweljakubas:
solves https://github.com/IntersectMBO/cardano-addresses/issues/321


angerman:
> hey @erikd @angerman in case we want notirization for macos, I created snippet that do it during releasing. What we need are setting the secrets connected with that. Maybe the team already has those established and set and we could reuse it here??? 

With official credentials this will be a lot harder. @johnalotoski would be the right person to talk to here. Due to the sensitivity involved and load on the SRE team, this isn't really something that can be solved quickly :-/

paweljakubas:
@angerman  sure! I will rebase the PR. Just wanted to prepare something that goes into this direction. I am leaving in your and Johns capable hands to prepare and "manage" secrets if there is appetite to support the macos notorization

PR #395, opened 2026-04-13
WASM target: compile cardano-addresses for wasm32-wasi

paolino:
## Summary

Enable `cardano-addresses` to compile to WebAssembly via GHC's WASM backend and export the binary as a reproducible Nix flake output. The WASM binary runs the same Haskell code as the native CLI, so the browser integration uses the real implementation rather than a separate JS crypto stack.

```bash
nix build github:IntersectMBO/cardano-addresses#wasm
ls result/cardano-addresses.wasm
```

## Problem

Two issues block compiling `cardano-addresses` to WASM:

1. The current `cardano-crypto` dependency chain (`basement` / `foundation` / `memory`) is archived, not compatible with GHC 9.12, and has no WASM support.
2. Building the WASM target reproducibly through Nix requires pinning the package indices and separating dependency fetching from the final offline build.

### 1. Vendor the wallet and BIP39 code actually used by `cardano-addresses`

This replaces the `basement` / `foundation` / `memory` dependency chain with the specific modules the project needs:

- Wallet cluster: `lib/Cardano/Address/Crypto/Wallet.hs`, `Wallet/Encrypted.hs`, `Wallet/Types.hs`, plus the required C sources in `cbits/ed25519/`
- BIP39 cluster: `lib/Cardano/Address/Crypto/BIP39.hs`, `BIP39/Dictionary.hs`, `BIP39/English.hs`

These are re-exported through `lib/Cardano/Address/Crypto.hs`.

### 2. Add WASM build support

`cabal-wasm.project` sets up the cross-compilation and applies the required package adjustments:

- `cborg` patched from git HEAD for the GHC 9.12 `IntWord64` fix
- `ram` built with WASI mmap emulation flags
- `crypton >= 1.1` uses `ram` instead of `memory`
- `formatting` built with `+no-double-conversion`
- `flag(wasm)` disables the native CLI executable and test suite in the WASM dependency set

### 3. Add a unified WASM executable

`browser/Main.hs` builds a single `cardano-addresses-wasm` executable that accepts JSON commands over WASI stdin/stdout.

Supported commands:

- `inspect`
- `derive`
- `make-address`
- `sign`
- `verify`
- `bootstrap-address`

### 4. Export a reproducible flake output

`nix/wasm.nix` uses a two-phase pattern:

1. Bootstrap a pinned local Hackage/CHaP view
2. Download dependencies into a deterministic store path
3. Build the WASM artifact offline from those pinned inputs

This makes the flake expose a reproducible `.#wasm` output while keeping the actual build logic in Nix.

### 5. CI

`.github/workflows/wasm.yml` builds the WASM target with `nix build .#wasm`, tests the resulting binary with `wasmtime`, and uploads the `.wasm` artifact.

## Files to review

- `nix/wasm.nix`
- `flake.nix`
- `browser/Main.hs`
- `cardano-addresses.cabal`
- `cabal-wasm.project`
- `README.md`
- `.github/workflows/wasm.yml`

## What this enables

This branch is used by `lambdasistemi/cardano-addresses-browser` so the browser app can run address inspection and key derivation against the real Haskell implementation compiled to WASM.


paweljakubas:
superb job! thank you @paolino

paweljakubas (approved):
🥇

PR #415, opened 2026-05-06
Support ghc-9.14

erikd:


paweljakubas (approved):
great, stuff! what about adding the this ghc version to workflow, ie. here https://github.com/IntersectMBO/cardano-addresses/blob/master/.github/workflows/haskell.yml#L15 ?

